Issue #45, opened 2018-03-25
Support for 3rd-party reporters?


autonomousapps:
See https://github.com/shyiko/ktlint/issues/176.

jeremymailen:
Would it work for you if I directly support https://github.com/mcassiano/ktlint-html-reporter (with the author's permission of course) as reporter for `'html'`?

I could make it extensible with providers the way rule sets are, but I suspect it is a less common use case and I'm shooting for "simply works out of the box" for the most part with this gradle plugin.

autonomousapps:
That would work for me, and I think it makes sense in this case. An html reporter would be super useful out of the box.

shyiko:
Guys, 
this is not stated in the ktlint's README (even though it should) but reporters are loaded using ServiceLoader same as rulesets. In other words, as long as jar containing Reporter implementation is on the classpath it will be loaded.
This is why something like

```groovy
dependencies {
    ktlint "com.github.shyiko:ktlint:0.20.0"
    ktlint "com.github.mcassiano:ktlint-html-reporter:master-SNAPSHOT"
}

task ktlint(type: JavaExec, group: "verification") {
    description = "Check Kotlin code style."
    classpath = configurations.ktlint
    main = "com.github.shyiko.ktlint.Main"
    args "--reporter=html", "src/**/*.kt" # notice that no artifact=... is needed
}
```
should work (same applies to Maven). By extension, https://github.com/jeremymailen/kotlinter-gradle#custom-rules can be used for the same purpose.
This is for your information only. I'm not advocating any particular implementation (or a lack thereof).

mcassiano:
Hey! I was going to publish this later on but there were some things I was trying to investigate further.

I'm not sure I was doing something wrong but loading the reporter from jar was not working. My guess is that in [line 378](https://github.com/shyiko/ktlint/blob/f5ec14f7e154741b581bbf9160a6d9e9c490f954/ktlint/src/main/kotlin/com/github/shyiko/ktlint/Main.kt#L378) the wrong id is being used to query the map (it's using the id passed on as an argument, which in this case is a path), when it should be using `ReporterProvider.id`. I didn't have time to check if that's what's happening but I'd gladly open an issue and PR to ktlint if that's the case. (ping @shyiko)

[...]

PR #46, opened 2018-03-29
Add html reporter

jeremymailen:
Alternative approach would be to allow extensible reporters similar to rulesets.

Addresses #45


jeremymailen:
@mcassiano let me know what you think. One question I have though is if this will require anyone applying this plugin to their gradle build to add jitpack to their `buildscript` repositories 🤔

mcassiano:
Looks good! I didn't know you'd need to use it directly like this. I think I can arrange publication to jcenter!

autonomousapps:
Excited to see this merged and the next version of kotlinter-gradle published! Thanks, guys :)

mcassiano:
I requested for ktlint-html-reporter to be included in jCenter! Now we wait

jeremymailen:
Perfect, thank you!

I didn't _need_ to use it directly this way. I could have made it extensible, but I think this will be easier for most users. Might still go the other way in the future if there's demand for it.

PR #54, opened 2018-06-01
Support sourceSet customization and rev to 2.0.0

### LINKED ISSUE #53: Plugin doesn't seem to take account changes to source set after plugin is applied
I'm using this plugin like this:

```
apply plugin: "com.android.application"
apply plugin: "kotlin-android"
apply plugin: "org.jmailen.kotlinter"

...
android {
    ...
    sourceSets {
        main {
            java.srcDirs += "src/main/kotlin"
        }
        test {
            java.srcDirs += "src/test/kotlin"
        }
        androidTest {
            java.srcDirs += "src/androidTest/kotlin"
        }
    }
}
```

The problem is kotlinter seems to resolve the source sets and collect the files right away, in the apply step. Changes made later have no effect. I'd like to use "src/group/kotlin" directories since these are pure kotlin projects.

What I'd like to see is the source file discovery deferred until task run time. I think it'd be fine if this plugin still searched for source sets on apply for purpose of creating the tasks, but I think each task should get a list of files from the app at runtime, rather than plugin apply time.

This is a fine plugin all around and I'd like to keep using it. I spent about an hour looking at how to hack a solution in, but it required a decent amount of refactoring and I don't have the time at the moment. I may pick it up in a few weeks when I have some spare time, but for now I'm leaving this issue open to put it on the radar and perhaps see if someone has a quick fix for it.

jeremymailen:
Fixes #53 

Before the build lifecycle was implemented such that if sourceSets were customized,
kotlinter would have already created tasks based on the original sourceSets. If we
do this instead during `project.afterEvaluate` it can work off of the final definitions.

One consequence is that any customization of kotlinter created tasks must be wrapped in an
`afterEvaluate` block since they don't exist during the first build configuration pass. This
prompts a bump to 2.0 since it will be a breaking check for some builds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Available on the Gradle Plugins Portal: https://plugins.gradle.org/plugin/org.jm
 
 ```groovy
 plugins {
-    id 'org.jmailen.kotlinter' version '1.12.0'
+    id 'org.jmailen.kotlinter' version '2.0.0'
 }
 ```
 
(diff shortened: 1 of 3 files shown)


jeremymailen:
closing in favor of #55

PR #55, opened 2018-06-04
Fix detection of customized SourceSets

### LINKED ISSUE #53: Plugin doesn't seem to take account changes to source set after plugin is applied

jeremymailen:
Before the plugin only created tasks correctly for source sets built into the convention
of the plugin (JVM or Android). Now we use the SourceSetContainer's `all { action }` block to create tasks in
a deferred manner after all have been defined.

It turns out this can be done without changing the outward behavior of the plugin, types of
tasks, or compatibility with older builds. It did require some substantial internal refactoring,
mostly to allow for each plugin source set resolver to take an action to apply for each
source set discovered which contains Kotlin files.

Fixes #53

### DIFF
--- a/src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
+++ b/src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
@@ -1,103 +1,83 @@
 package org.jmailen.gradle.kotlinter
 
 import com.android.build.gradle.BaseExtension
+import com.android.build.gradle.api.AndroidSourceSet
 import org.gradle.api.Plugin
 import org.gradle.api.Project
-import org.gradle.api.file.FileCollection
 import org.gradle.api.file.FileTree
 import org.gradle.api.file.SourceDirectorySet
 import org.gradle.api.internal.HasConvention
 import org.gradle.api.plugins.JavaPluginConvention
 import org.gradle.api.tasks.SourceSet
+import org.gradle.api.tasks.SourceSetContainer
 import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet
 import org.jmailen.gradle.kotlinter.support.reporterFileExtension
 import org.jmailen.gradle.kotlinter.tasks.FormatTask
 import org.jmailen.gradle.kotlinter.tasks.LintTask
+import java.io.File
 
 class KotlinterPlugin : Plugin<Project> {
 
     val extendablePlugins = mapOf(
-            "org.jetbrains.kotlin.jvm" to this::kotlinSourceSets,
-            "kotlin-android" to this::androidSourceSets)
+            "org.jetbrains.kotlin.jvm" to KotlinJvmSourceSetResolver,
+            "kotlin-android" to AndroidSourceSetResolver)
 
     override fun apply(project: Project) {
-        val kotlinterExtention = project.extensions.create("kotlinter", KotlinterExtension::class.java)
+        val kotlinterExtension = project.extensions.create("kotlinter", KotlinterExtension::class.java)
 
         // for known kotlin plugins, create tasks by convention.
-        val kotlinApplier = KotlinterApplier(project)
+        val taskCreator = TaskCreator(project)
         extendablePlugins.forEach { pluginId, sourceResolver ->
             project.plugins.withId(pluginId) {
-                val sourceSets = sourceResolver(project)
-                kotlinApplier.createTasks(sourceSets)
+
+                sourceResolver.applyToAll(project) { id, files ->
+                    taskCreator.createSourceSetTasks(id, files)
+                }
             }
         }
+        taskCreator.createParentTasks()
 
         project.afterEvaluate {
-            kotlinApplier.lintTasks.forEach { lintTask ->
-                lintTask.ignoreFailures = kotlinterExtention.ignoreFailures
-                lintTask.indentSize = kotlinterExtention.indentSize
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
```diff
@@ -1,104 +1,84 @@
...
 
         // for known kotlin plugins, create tasks by convention.
-        val kotlinApplier = KotlinterApplier(project)
+        val taskCreator = TaskCreator(project)
         extendablePlugins.forEach { pluginId, sourceResolver ->
             project.plugins.withId(pluginId) {
-                val sourceSets = sourceResolver(project)
-                kotlinApplier.createTasks(sourceSets)
+
+                sourceResolver.applyToAll(project) { id, files ->
```
mikeholler: It looks like source sets are resolved at application time, which means if a source set is added after this plugin is applied it won't pick up on it. I would consider using SourceSet's `whenObjectAdded { ... }` function to allow you to pick up source sets that are created long after this plugin is applied.

src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
```diff
@@ -1,104 +1,84 @@
...
+                    taskCreator.createSourceSetTasks(id, files)
+                }
             }
         }
+        taskCreator.createParentTasks()
 
         project.afterEvaluate {
```
mikeholler: So much of this refactor depends on this decision right here.  As I mentioned before in the linked issue, `afterEvaluate` is a bit of an anti-pattern that should be avoided if possible. It creates builds that are more difficult to understand, because it's hard for the user to know where these tasks are created if they don't exist at the same time a source set does. That being said, keeping `afterEvaluate` allows you to keep your existing API without breaking backwards compatibility. I think that'd be hard to do if you tried to get rid of `afterEvaluate`.

What sort of direction do you ultimately see this library going? If you were writing this from scratch today, would you still want to extend SourceTask to allow for user-modification of files crawled by these tasks, or would you rather a task was assigned a SourceSet and that was the only thing about the input files that you could modify?

If you see it moving away from `afterEvaluate`, what's stopping the shift now? Development effort and backwards compatibility are probably good (and respectable) answers for putting off that change until later, but beyond that I can't see any more.

src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
mikeholler: I _think_ this would also be a pretty easy change that would not break any backwards compatibility (unlike my other comment).

jeremymailen: This refactor is subtle, but we no longer resolve source sets at apply time. We do it on the source set `all` as you wisely suggested we do.

Take a close read of `applyToAll` in the two implemenations `KotlinJvmSourceSetResolver` and `AndroidSourceSetResolver`. The second argument of `applyToAll` is a block to be called within `SourceSetContainer.all { ... }`. So the `TaskCreator` will do its thing once the source sets are resolved properly for the project.

src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
jeremymailen: At this point, the only thing we are doing in `afterEvaluate` is applying the configuration from the
```Groovy
kotlinter {
  <config>
}
```
extension to the tasks. There may well be a better way to make sure that the extension properties get applied to the tasks already created, but for the moment it seems relatively harmless?

It turned out that I didn't need to create different types of tasks (`SourceSetTask` vs `SourceTask`). It was perfectly valid to create a `SourceTask` within the `SourceSetContainer.all` and not have to break the existing API -- best of both worlds!

jeremymailen:
Take a 2nd look and if you can, test it out on your Android project. Suggested testing includes:
- a format and lint task exists for each of your custom source sets (and any standard source sets you expect)
- `gradle lintKotlin` processes all of the source set lint tasks
- `gradle formatKotlin` processes all of the source set format tasks

thanks again for the review!

mikeholler:
I'm working on tests in #57. It's not done yet though so don't merge.

PR #73, opened 2018-09-26
Support relocatable build cache for lint, remove incremental support …

jeremymailen:
…from format

For format incremental build support is rarely useful and there's an edge case where it won't re-run if you revert changes to prior state.


jeremymailen:
Hopefully fully addresses issues raised in #77

PR #75, opened 2018-10-04
Upgrade to ktlint 0.29.0 and fix warnings on incremental build properties

jeremymailen:


src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/FormatTask.kt
```diff
@@ -13,6 +14,7 @@ import java.io.File
 
 open class FormatTask : SourceTask() {
 
+    @Internal
```
jeremymailen: We do this to prevent incremental build, which doesn't work quite right if you revert after since this task modifies input files.

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/LintTask.kt
```diff
@@ -26,6 +27,10 @@ open class LintTask : SourceTask() {
     @PathSensitive(PathSensitivity.RELATIVE)
     lateinit var reports: Map<String, File>
 
+    @InputFiles
+    @PathSensitive(PathSensitivity.RELATIVE)
+    override fun getSource() = super.getSource()
```
jeremymailen: The desirable path sensitivity does not get inherited from `SourceTask` so it looks like we need to override just to annotate it.

PR #83, opened 2019-02-22
Added two functional tests in preparation for Worker migration.

autonomousapps:
In preparation for use of the Worker API, I wanted to ensure there was some test coverage of the features in question, first. These tests use Gradle TestKit. I had to make a few adjustments to `build.gradle`, the most significant of which was
```
implementation 'org.jetbrains.kotlin:kotlin-gradle-plugin'
```
Without that, the functional tests wouldn't compile, since some required classes were not on the test compile classpath. Something to do with how TestKit works.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,15 +16,17 @@ repositories {
 dependencies {
     implementation 'com.github.shyiko:ktlint:0.30.0'
     implementation 'me.cassiano:ktlint-html-reporter:0.2.0'
-    implementation 'org.jetbrains.kotlin:kotlin-gradle-plugin'
     
-    compileOnly 'com.android.tools.build:gradle:3.0.1' // TODO this is a different version than the testRuntime dependency
-    compileOnly('org.jetbrains.kotlin:kotlin-reflect') {
-        because 'AGP has a transitive dependency on this, and we want to pin to a particular version'
-    }
+    compileOnly 'org.jetbrains.kotlin:kotlin-gradle-plugin'
+    compileOnly 'com.android.tools.build:gradle:3.3.1'
 
     testImplementation 'junit:junit:4.12'
-    testRuntime 'com.android.tools.build:gradle:3.1.3'
+    testRuntime 'com.android.tools.build:gradle:3.3.1'
+}
+
+// Required to put the Kotlin plugin on the classpath for the functional test suite
+tasks.withType(PluginUnderTestMetadata).configureEach {
+    pluginClasspath.from(configurations.compileOnly)
 }
 
 version = '1.21.0'
--- a/build.gradle
+++ b/build.gradle
@@ -14,12 +14,16 @@ repositories {
 }
 
 dependencies {
-    compile 'com.github.shyiko:ktlint:0.30.0'
-    compile 'me.cassiano:ktlint-html-reporter:0.2.0'
-    compileOnly 'org.jetbrains.kotlin:kotlin-gradle-plugin'
-    compileOnly 'com.android.tools.build:gradle:3.0.1'
-    testCompile 'junit:junit:4.12'
-    testRuntime 'org.jetbrains.kotlin:kotlin-gradle-plugin'
+    implementation 'com.github.shyiko:ktlint:0.30.0'
+    implementation 'me.cassiano:ktlint-html-reporter:0.2.0'
+    implementation 'org.jetbrains.kotlin:kotlin-gradle-plugin'
+    
+    compileOnly 'com.android.tools.build:gradle:3.0.1' // TODO this is a different version than the testRuntime dependency
+    compileOnly('org.jetbrains.kotlin:kotlin-reflect') {
+        because 'AGP has a transitive dependency on this, and we want to pin to a particular version'
+    }
+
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -14,12 +14,16 @@ repositories {
...
-    compile 'me.cassiano:ktlint-html-reporter:0.2.0'
-    compileOnly 'org.jetbrains.kotlin:kotlin-gradle-plugin'
-    compileOnly 'com.android.tools.build:gradle:3.0.1'
-    testCompile 'junit:junit:4.12'
-    testRuntime 'org.jetbrains.kotlin:kotlin-gradle-plugin'
+    implementation 'com.github.shyiko:ktlint:0.30.0'
+    implementation 'me.cassiano:ktlint-html-reporter:0.2.0'
+    implementation 'org.jetbrains.kotlin:kotlin-gradle-plugin'
+    
+    compileOnly 'com.android.tools.build:gradle:3.0.1' // TODO this is a different version than the testRuntime dependency
```
jeremymailen: you can safely bump this to `3.1.3`, I tested and it worked.

build.gradle
```diff
@@ -14,12 +14,16 @@ repositories {
...
 dependencies {
-    compile 'com.github.shyiko:ktlint:0.30.0'
-    compile 'me.cassiano:ktlint-html-reporter:0.2.0'
-    compileOnly 'org.jetbrains.kotlin:kotlin-gradle-plugin'
-    compileOnly 'com.android.tools.build:gradle:3.0.1'
-    testCompile 'junit:junit:4.12'
-    testRuntime 'org.jetbrains.kotlin:kotlin-gradle-plugin'
+    implementation 'com.github.shyiko:ktlint:0.30.0'
+    implementation 'me.cassiano:ktlint-html-reporter:0.2.0'
+    implementation 'org.jetbrains.kotlin:kotlin-gradle-plugin'
```
jeremymailen: I couldn't get it to work either with a lesser scope. Even putting it both in `compileOnly` and `testImplementation`.

My original reason for using `compileOnly` for all the gradle dependencies was I was worried about it leaking into the buildscript dependencies and changing the version of gradle plugins the end-user was trying to use.

Even with `implementation` it does seem to introduce a change into the results of `gradle buildEnvironment` for a project using this plugin:
```
classpath
+--- org.jmailen.gradle:kotlinter-gradle:1.21.0
|    +--- com.github.shyiko:ktlint:0.30.0
|    |    +--- org.jetbrains.kotlin:kotlin-stdlib:1.3.20 -> 1.3.21
vvv newly introduced dependency
...
|    \--- org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.20 -> 1.3.21
|         +--- org.jetbrains.kotlin:kotlin-gradle-plugin-api:1.3.21
|         |    +--- org.jetbrains.kotlin:kotlin-stdlib:1.3.21 (*)
...
^^^ newly introduced dependency
\--- org.jetbrains.kotlin.jvm:org.jetbrains.kotlin.jvm.gradle.plugin:1.3.21
     \--- org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.21 (*)
```

What do you think? It seems like it could bump the user's kotlin gradle plugin up a version unintentionally in certain circumstances?

autonomousapps: I'll experiment a bit and see what our options are. I agree it's worth the time to get a better understanding.

autonomousapps: I admit I've been unable to figure out why it seems to need that dependency in the `implementation` configuration. I was looking at [your competitor](https://github.com/JLLeitschuh/ktlint-gradle/blob/master/plugin/build.gradle.kts) ( ;-) ), and he is using `compileOnly` and `testImplementation`.

jeremymailen: I believe ktlint-gradle executes the ktlint tool as a javaexec for each task whereas kotlinter-gradle runs ktlint's underlying engine inside the gradle runtime. This might explain the dependency differences and how they might affect TestKit.

jeremymailen: I can describe what I do to test. I know it's manual and tedious, but it's otherwise hard to validate the plugin actually works on different project types given it's relationship both with the kotlin compiler and the kotlin jvm and android plugins it extends.

Change the kotlinter-gradle version to be my branch name.
`gradle publishToMavenLocal`

I have a couple of projects that use kotlinter-gradle, so I change the build.gradle to have:
```groovy
buildscript {
    repositories {
        mavenLocal()
        jcenter()
    }

    dependencies {
        classpath 'org.jmailen.gradle:kotlinter-gradle:my-branch-name'
    }
}
```

On the project that uses it, I run through any fresh and incremental linting and formatting tests I want to do. It's not as terrible as it sounds once you get it set up the first time.

autonomousapps: I was able to resolve the issue (with inspiration from [here](https://github.com/JLLeitschuh/ktlint-gradle/blob/master/plugin/build.gradle.kts#L18-L20)).

autonomousapps:
I bumped AGP to 3.3.1 because
1. It was latest stable
2. It resolved some warnings
3. We no longer need to explicitly depend on kotlin-reflect.

PR #86, opened 2019-03-28
Uses a parallelStream() to substantially improve performance on CPUs …

samzurcher:
…with many cores

Not so nicely visible in the diff but the only change I have made is to add 

```
.toList()
.parallelStream()
```

plus added indentation.

This change helps run the checks substantially faster on a multi-core CPU since multiple threads are started by parallelStream(). This is possible since the linting can be nicely done in parallel.

### DIFF
--- a/src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/LintTask.kt
+++ b/src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/LintTask.kt
@@ -55,31 +55,34 @@ open class LintTask : SourceTask() {
 
         fileReporters.onEach { it.beforeAll() }
 
-        getSource().forEach { file ->
-            val relativePath = file.toRelativeString(project.projectDir)
-            fileReporters.onEach { it.before(relativePath) }
-            logger.log(LogLevel.DEBUG, "$name linting: $relativePath")
+        getSource()
+            .toList()
+            .parallelStream()
+            .forEach { file ->
+                val relativePath = file.toRelativeString(project.projectDir)
+                fileReporters.onEach { it.before(relativePath) }
+                logger.log(LogLevel.DEBUG, "$name linting: $relativePath")
 
-            val lintFunc = when (file.extension) {
-                "kt" -> this::lintKt
-                "kts" -> this::lintKts
-                else -> {
-                    logger.log(LogLevel.DEBUG, "$name ignoring non Kotlin file: $relativePath")
-                    null
+                val lintFunc = when (file.extension) {
+                    "kt" -> this::lintKt
+                    "kts" -> this::lintKts
+                    else -> {
+                        logger.log(LogLevel.DEBUG, "$name ignoring non Kotlin file: $relativePath")
+                        null
+                    }
                 }
+
+                lintFunc?.invoke(file, resolveRuleSets(experimentalRules)) { error ->
+                    fileReporters.onEach { it.onLintError(relativePath, error, false) }
+
+                    val errorStr = "$relativePath:${error.line}:${error.col}: ${error.detail}"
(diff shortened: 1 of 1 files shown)


autonomousapps:
I would hesitate to do this for a number of reasons.
1. Using non-gradle-managed threads means we lose log output
1. This could exceed the maximum amount of work that is set for the build. On a small project, it might not be a big deal to spawn, say, 10 threads. But on a larger project, you'll have NumModules * 10 (or whatever), and now it's a problem. You're going to prevent gradle from running other tasks in parallel.
1. I have no idea if Reporters are thread-safe.

PS: We need to update the functional tests to catch potential issues like these.

jeremymailen:
First off, @samzurcher thank you so much for the contribution! I love when people find potential improvements _and_ do the work to make them happen.

These are some valid concerns @autonomousapps. With reporters, I don't know about thread safety either. At a minimum it might report results in a different order each time, which may or may not impact anyone.

Is using the Gradle Worker API the right solution here? Perhaps it would need to be at the source set (task) level rather than per file if report generation is an obstacle. If so, I imagine formatting could be parallelized too?

samzurcher:
@autonomousapps Ok, I see your points:
1. Ok, better to do this with https://guides.gradle.org/using-the-worker-api/ (as @jeremymailen has mentioned)
2. Using the worker api from 1. should solve this issue too; agreed?
3. Ok, I would need to investigate that or even better make a thread-safe wrapper for the reporters.

@jeremymailen Do you know of a use-case in which the order of the reports matter? If so, we could just collect the reports and write them in a sorted way as soon as the processing has finished. This adds the disadvantage though that the reports will be written only at the end.
To be honest, I very much prefer doing it at the file level instead of the source set level as this will allow for much greater use of the available cores.

And agreed, formatting could be parallelized too. 

Please decline this pull request. You'll get a new one based on above (and potentially following) discussions one on Monday.

jeremymailen:
I use the reporters from ktlint and it does look like they are thread safe when it comes to logging each error and starting/ending the processing of the file:
https://github.com/pinterest/ktlint/blob/master/ktlint-core/src/main/kotlin/com/github/shyiko/ktlint/core/Reporter.kt#L4

I don't know if any use cases for the reports being ordered in a certain way, although I could imagine non-stable results could be confusing. I haven't looked at the reporter implementations in detail, but at a quick glance it seems they do some post work to group the results, so all _might_ be well there.

jeremymailen:
Thank you and I look forward to your new implementation @samzurcher!

autonomousapps:
Using the Worker API would be preferred, but it still relies on the reporters being thread-safe (depending on how you implement your solution, see below). There are two ways to do it:
1. Push all the work, together (i.e., the entire `source` input) into a single work item. This is still better than the current situation, because now Gradle can schedule other tasks within that same project concurrently. (In general, tasks take a lock on a `Project` instance when they begin execution, preventing other tasks in that same project from executing.) With this scenario, I don't think we need to worry much about thread-safety, since all the "work" is still done in a single thread.
1. Try to push every file into a separate work item. This should be much faster, but it's more complex to implement, and now thread-safety matters.

And of course we can imagine doing (1) first, and then migrating to (2) later.

The order of reporters would matter if the output order changed between runs. This would have implications for the cacheability of the task. So long as the ordering is eventually sorted, I think it's fine.

autonomousapps:
As a follow-up to my earlier PS, I think that the functional tests should be updated to include a "multi-module" build. This will provide some guarantees that the Worker migration doesn't break anything. I'm open to working on this, but I don't know when I'll have time.

jeremymailen:
Closing in favor of https://github.com/jeremymailen/kotlinter-gradle/pull/87
Thank you!

PR #89, opened 2019-04-05
Parallelized Formatting using the Gradle Worker API

samzurcher:
In the same tradition as https://github.com/jeremymailen/kotlinter-gradle/pull/87, parallelizing the formatting with the Gradle Worker API.

Comments / Questions / Open Points
1. I have used a lower default batch size for formatting (20) compared to linting (30). At least on my setup, this seems to perform better. I see speedup from 12.6s to 1.8s.
2. I'm not 100% about the Gradle Isolation Mode (see `FormatWorkerConfigurationAction`) but think that we can still use `IsolationMode.NONE`.

Feedback welcome!

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,7 +216,6 @@ Reporters behave as described at: https://github.com/shyiko/ktlint
 The `experimentalRules` property enables rules which are part of ktlint's experimental rule set.
 
 The `fileBatchSize` property configures the number of files that are processed in one Gradle Worker API call.
-Default value is 30 for linting and 20 for formatting.
 
 ### Customizing Tasks
 
--- a/README.md
+++ b/README.md
@@ -216,6 +216,7 @@ Reporters behave as described at: https://github.com/shyiko/ktlint
 The `experimentalRules` property enables rules which are part of ktlint's experimental rule set.
 
 The `fileBatchSize` property configures the number of files that are processed in one Gradle Worker API call.
+Default value is 30 for linting and 20 for formatting.
 
 ### Customizing Tasks
 
--- a/src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/format/FormatExecutionContext.kt
+++ b/src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/format/FormatExecutionContext.kt
@@ -2,11 +2,13 @@ package org.jmailen.gradle.kotlinter.tasks.format
 
 import org.gradle.api.logging.Logger
 import org.jmailen.gradle.kotlinter.support.ExecutionContext
+import java.util.Queue
+import java.util.concurrent.ConcurrentLinkedQueue
 
 /**
  * Execution context for the formatting task.
  */
 data class FormatExecutionContext(
     override val logger: Logger,
-    val fixes: MutableList<String> = mutableListOf()
+    val fixes: Queue<String> = ConcurrentLinkedQueue<String>()
 ) : ExecutionContext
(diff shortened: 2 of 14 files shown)


README.md
```diff
@@ -216,6 +216,7 @@ Reporters behave as described at: https://github.com/shyiko/ktlint
 The `experimentalRules` property enables rules which are part of ktlint's experimental rule set.
 
 The `fileBatchSize` property configures the number of files that are processed in one Gradle Worker API call.
+Default value is 30 for linting and 20 for formatting.
```
jeremymailen: It is a bit strange to have one config property that changes two different defaults. Do you think the needs are different enough they need two separate values? I'd be in favor or one param and one default if there's a reasonable shared value that's close enough to optimal for both linting and formatting -- mostly for the sake of simplicity since I doubt most people will need to tune it to that level of detail.

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/format/FormatExecutionContext.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import org.gradle.api.logging.Logger
+import org.jmailen.gradle.kotlinter.support.ExecutionContext
+
+/**
+ * Execution context for the formatting task.
+ */
+data class FormatExecutionContext(
+    override val logger: Logger,
+    val fixes: MutableList<String> = mutableListOf()
```
jeremymailen: Does this collection need to be threadsafe?

samzurcher: Yes, I realized that after I pushed the changes - you'll get a fix on Monday...

README.md
samzurcher: Ok, i'll change it back to one default. And if needed, people can still set the value explicitely.

autonomousapps:
It's not 100% clear that we would want formatting to be Workerized. One of the consequences is that other tasks _within the same `Project`_, could run simultaneously. However, since you're actually modifying source code, maybe that's not a good thing. Something to consider.

jeremymailen:
Valid concern @autonomousapps. The interesting part here is that `lintKotlin` isn't hooked into any part of the build lifecycle as a dependency. Anyone who does add it as a dependency to say, `compileKotlin` would need to make sure any potentially conflicting tasks are explicitly dependent on or dependencies of `lintKotlin`. Although it could be a gotcha, I think it's ok. Speeding up formatting could make things better for someone who wants to implement a "watch" style workflow that automatically fixes their code. It's probably worth testing that setup to validate it doesn't run into any conflicts.

autonomousapps:
@jeremymailen that makes sense. I think some documentation might be in order, to discourage potential bad patterns. In my own projects, I don't include `formatKotlin` as part of any automated workflows. I just run it when there's a lint error. On CI, I run `lintKotlin` (or is it `kotlinLint`?) and fail the build if there's an error, and let the dev fix how s/he likes.

jeremymailen:
I haven't forgotten this @samzurcher! There were a couple of bugs I wanted to fix before getting this incorporated and released, stay tuned. Just waiting on the kotlin 1.3.30 compatibility to get resolved since that's probably a requirement for anyone to upgrade and get the new feature.

samzurcher:
No worries and no hurry. We are not using the formatting part in our application; so feel free to take your time!

jeremymailen:
Will commit the rebased one which takes the package moves into account. Thanks @samzurcher!

PR #124, opened 2020-01-15
Fix #115 by lazy task sources evaluation

### LINKED ISSUE #115: FormatKotlin task doesn't seem to format code in /kotlin sourceset.
FormatKotlin task doesn't seem to work for code inside /src/main/kotlin/* sourcesets. But it works just fine for /src/main/java/*.

mateuszkwiecinski:
Fixes #115

I let myself not only to lazily evaluate sources, but also to enable lazy task configuration.

I hope that addresses: 
>I think it's fixable if we use a gradle Provider or Closure to provide the final srcDirs as use time instead of configuration time

mentioned https://github.com/jeremymailen/kotlinter-gradle/issues/115#issuecomment-522457021

However, when delaying sources configuration we have to create tasks that weren't previously created when source set was empty. So that would be slight behavior change and I couldn't figure out a way of not creating those. Although, that shouldn't be an issue as those tasks are `NO_SOURCE` so they are not executed.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+[*.{kt,kts}]
+max_line_length = 140
+indent_size = 4
+continuation_indent_size = 4
+insert_final_newline = true
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-    id 'org.jetbrains.kotlin.jvm' version '1.3.60'
+    id 'org.jetbrains.kotlin.jvm' version '1.3.61'
     id 'com.gradle.plugin-publish' version '0.10.1'
     id 'java-gradle-plugin'
     id 'maven-publish'
@@ -8,7 +8,6 @@ plugins {
 }
 
 repositories {
-    mavenLocal()
     jcenter()
     google()
 }
@@ -25,11 +24,12 @@ dependencies {
     implementation "com.pinterest.ktlint:ktlint-ruleset-standard:$ktlintVers"
 
     compileOnly 'org.jetbrains.kotlin:kotlin-gradle-plugin'
-    compileOnly 'com.android.tools.build:gradle:3.4.2'
+    compileOnly 'com.android.tools.build:gradle:3.5.3'
 
-    testImplementation 'junit:junit:4.12'
-    testImplementation 'com.nhaarman.mockitokotlin2:mockito-kotlin:2.1.0'
-    testRuntime 'com.android.tools.build:gradle:3.4.2'
+    testImplementation 'junit:junit:4.13'
+    testImplementation 'com.nhaarman.mockitokotlin2:mockito-kotlin:2.2.0'
+    testImplementation 'org.jetbrains:annotations:18.0.0'
+    testRuntime 'com.android.tools.build:gradle:3.5.3'
 }
 
 // Required to put the Kotlin plugin on the classpath for the functional test suite
@@ -62,6 +62,10 @@ pluginBundle {
             description = 'Lint and formatting for Kotlin using ktlint with configuration-free setup on JVM and Android projects'
         }
     }
+}
+
+java {
+    withSourcesJar()
 }
 
 publishing {
--- a/build.gradle
+++ b/build.gradle
@@ -29,8 +29,6 @@ dependencies {
     testImplementation 'junit:junit:4.13'
     testImplementation 'com.nhaarman.mockitokotlin2:mockito-kotlin:2.2.0'
     testImplementation 'org.jetbrains:annotations:18.0.0'
-    testRuntime 'com.android.tools.build:gradle:3.5.3'
-    testRuntime 'org.jetbrains.kotlin:kotlin-gradle-plugin'
 }
 
 // Required to put the Kotlin plugin on the classpath for the functional test suite
--- a/build.gradle
+++ b/build.gradle
@@ -30,6 +30,7 @@ dependencies {
     testImplementation 'com.nhaarman.mockitokotlin2:mockito-kotlin:2.2.0'
     testImplementation 'org.jetbrains:annotations:18.0.0'
     testRuntime 'com.android.tools.build:gradle:3.5.3'
+    testRuntime 'org.jetbrains.kotlin:kotlin-gradle-plugin'
 }
 
 // Required to put the Kotlin plugin on the classpath for the functional test suite
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-    id 'org.jetbrains.kotlin.jvm' version '1.3.61'
+    id 'org.jetbrains.kotlin.jvm' version '1.3.60'
     id 'com.gradle.plugin-publish' version '0.10.1'
     id 'java-gradle-plugin'
     id 'maven-publish'
@@ -31,7 +31,7 @@ dependencies {
     testImplementation 'org.jetbrains:annotations:18.0.0'
 }
 
-// Required to put the Kotlin plugin on the classpath for the functional test suite
+// Required to put all `compileOnly` dependencies on the classpath for the functional test suite
 tasks.withType(PluginUnderTestMetadata).configureEach {
     pluginClasspath.from(configurations.compileOnly)
 }
--- a/src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
+++ b/src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
@@ -42,7 +42,7 @@ class KotlinterPlugin : Plugin<Project> {
                         lintTask.reports = kotlinterExtension.reporters.associate { reporter ->
                             reporter to reportFile("$id-lint.${reporterFileExtension(reporter)}")
                         }
-                        lintTask.ktLintParams = kotlinterExtension.ktlintParams(editorConfigPath())
+                        lintTask.ktLintParams = kotlinterExtension.toKtLintParams(editorConfigPath())
                         lintTask.fileBatchSize = kotlinterExtension.fileBatchSize
                     }
                     lintKotlin.dependsOn(lintKotlinPerVariant)
@@ -50,7 +50,7 @@ class KotlinterPlugin : Plugin<Project> {
                     val formatKotlinPerVariant = tasks.register("formatKotlin${id.capitalize()}", FormatTask::class.java) { formatTask ->
                         formatTask.source(resolveSources)
                         formatTask.report = reportFile("$id-format.txt")
-                        formatTask.ktLintParams = kotlinterExtension.ktlintParams(editorConfigPath())
+                        formatTask.ktLintParams = kotlinterExtension.toKtLintParams(editorConfigPath())
                         formatTask.fileBatchSize = kotlinterExtension.fileBatchSize
                     }
                     formatKotlin.dependsOn(formatKotlinPerVariant)
@@ -119,7 +119,7 @@ internal object AndroidSourceSetResolver : SourceSetResolver {
     }
 }
 
-private fun KotlinterExtension.ktlintParams(editorConfigPath: String?) = KtLintParams(
+private fun KotlinterExtension.toKtLintParams(editorConfigPath: String?) = KtLintParams(
     indentSize,
     continuationIndentSize,
     experimentalRules,
--- a/src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
+++ b/src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
@@ -28,7 +28,7 @@ class KotlinterPlugin : Plugin<Project> {
     override fun apply(project: Project) = with(project) {
         val kotlinterExtension = extensions.create("kotlinter", KotlinterExtension::class.java)
 
-        // for known kotlin plugins, create tasks by convention.
+        // for known kotlin plugins, register tasks by convention.
         extendablePlugins.forEach { (pluginId, sourceResolver) ->
             plugins.withId(pluginId) {
 
--- a/src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
+++ b/src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
@@ -36,21 +36,13 @@ class KotlinterPlugin : Plugin<Project> {
                 val formatKotlin = registerParentFormatTask()
 
                 sourceResolver.applyToAll(project) { id, resolveSources ->
-                    val ktLintParams = KtLintParams(
-                        kotlinterExtension.indentSize,
-                        kotlinterExtension.continuationIndentSize,
-                        kotlinterExtension.experimentalRules,
-                        kotlinterExtension.disabledRules,
-                        editorConfigPath()
-                    )
-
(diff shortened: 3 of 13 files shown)


.editorconfig
```diff
@@ -0,0 +1,5 @@
+[*.{kt,kts}]
+max_line_length = 140
+indent_size = 4
+continuation_indent_size = 4
+insert_final_newline = true
```
mateuszkwiecinski: My goal here was to standarize code style settings for the project, I can remove that or move to a separate PR if needed.

build.gradle
```diff
@@ -8,7 +8,6 @@ plugins {
 }
 
 repositories {
-    mavenLocal()
```
mateuszkwiecinski: I removed that to have more predictable builds. I guess no one here should use dependencies from there.

build.gradle
```diff
@@ -25,11 +24,13 @@ dependencies {
...
     compileOnly 'org.jetbrains.kotlin:kotlin-gradle-plugin'
-    compileOnly 'com.android.tools.build:gradle:3.4.2'
+    compileOnly 'com.android.tools.build:gradle:3.5.3'
 
-    testImplementation 'junit:junit:4.12'
-    testImplementation 'com.nhaarman.mockitokotlin2:mockito-kotlin:2.1.0'
-    testRuntime 'com.android.tools.build:gradle:3.4.2'
+    testImplementation 'junit:junit:4.13'
+    testImplementation 'com.nhaarman.mockitokotlin2:mockito-kotlin:2.2.0'
+    testImplementation 'org.jetbrains:annotations:18.0.0'
```
mateuszkwiecinski: I've bumped dependencies version and added 2 new dependencies
> org.jetbrains:annotations:18.0.0

For better IDE support (mainly the `@Language` annotation)

> org.jetbrains.kotlin:kotlin-gradle-plugin

As that's what I starter to need for tests I wrote

build.gradle
```diff
@@ -64,6 +65,10 @@ pluginBundle {
     }
 }
 
+java {
+    withSourcesJar()
```
mateuszkwiecinski: I thought it would be nice to publish sources too (`-sources.jar` artifact would be added to maven publicatioN)
image

src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
```diff
@@ -20,146 +21,117 @@ import org.jmailen.gradle.kotlinter.tasks.LintTask
...
+        project.fileTree(dir) { it.include("**/*.kt") }
     }.reduce { merged: FileTree, tree ->
-        merged.plus(tree)
-    }.files
+        merged + tree
+    }
 }
 
-val String.id
+internal val String.id
```
mateuszkwiecinski: I let myself to set visibility of those to `internal` to avoid exposing those as that was something I've noticed when using `Kotlinter` plugin. I assume those aren't meant to be used outside the library, right?

src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
```diff
@@ -20,146 +21,117 @@ import org.jmailen.gradle.kotlinter.tasks.LintTask
...
+                val formatKotlin = registerParentFormatTask()
+
+                sourceResolver.applyToAll(project) { id, resolveSources ->
+                    val ktLintParams = KtLintParams(
+                        kotlinterExtension.indentSize,
+                        kotlinterExtension.continuationIndentSize,
+                        kotlinterExtension.experimentalRules,
+                        kotlinterExtension.disabledRules,
+                        editorConfigPath()
+                    )
```
mateuszkwiecinski: Oh, I've started wondering if that won't be too quckly to read from extension. In case when tests don't cover such I'll make them verify that.

mateuszkwiecinski: Yup, that was too early. Tests didn't cover passing parameters using Kotlinter extension so I added them. 
I changed the code so the `KtLintParams` are obtained when tasks are configured.

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/ExtensionTest.kt
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+
+    @Test
+    fun `extension configures indentSize`() {
+        projectRoot.resolve("build.gradle") {
+            @Language("groovy")
+            val script = """
+                kotlinter {
+                    indentSize = 2
```
mateuszkwiecinski: I didn't have any good idea how to test if setting `continuationIndentSize`, `experimentalRules` or `fileBatchSize` works as expected.

build.gradle
jeremymailen: I believe I included this because my typical test flow involves `publishToMavenLocal` and testing the new version of the plugin on some real example kotlin backend and android projects I have locally. In any case, that's something I can edit in ad hoc.

jeremymailen:
Thank you for this great contribution and the notes! I'll give it a more thorough review and test it out this weekend, hopefully get it incorporated in a new release.

build.gradle
mateuszkwiecinski: I'm not sure I understood. `publishToMavenLocal` task doesn't rely on having `mavenLocal` repository. It only can be useful if you test locally published _Kotlinter_ on this project itself - does that take place? If it is, I'll revert that change - I didn't wan't to make anyone to change habbits.

I'm proposing to remove that as I couldn't see any use case here and basing on my experience and [Gradle documentation](https://docs.gradle.org/current/userguide/declaring_repositories.html#sec:case-for-maven-local) having `mavenLocal` may lead to unpredictible builds (especially when repositories order is one of the factors in depenedency resolution process)

build.gradle
mateuszkwiecinski: > As that's what I started to need for tests I wrote

I was wrong. All the `testRuntime` wasn't actually needed as all `compileOnly` dependencies are added to test runtime using
```
tasks.withType(PluginUnderTestMetadata).configureEach {
    pluginClasspath.from(configurations.compileOnly)
}
```
few lines below.

build.gradle
jmeekhof: My flow for testing plugins involves a separate repository, therefor I don't think you really need to have `mavenLocal` in the build file.

src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
```diff
@@ -2,13 +2,14 @@ package org.jmailen.gradle.kotlinter
 
 import com.android.build.gradle.BaseExtension
 import com.android.build.gradle.api.AndroidSourceSet
-import java.io.File
+import com.android.build.gradle.internal.tasks.factory.dependsOn
```
jeremymailen: This will cause non-android projects to fail. You'll need to use the generic dependency setting, see below

src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
```diff
@@ -20,146 +21,117 @@ import org.jmailen.gradle.kotlinter.tasks.LintTask
...
-        }
-
-        val lintKotlin = project.tasks.create("lintKotlin") {
+    private fun Project.registerParentLintTask() =
+        tasks.register("lintKotlin") {
             it.group = "formatting"
             it.description = "Runs lint on the Kotlin source files."
-            it.dependsOn(lintTasks)
+        }.also {
+            tasks.named("check").dependsOn(it)
```
jeremymailen: Once you remove the android dependency, you can do it generically like this:
```kotlin
tasks.named("check").configure { check -> check.dependsOn(it) }
```

build.gradle
```diff
@@ -1,5 +1,5 @@
 plugins {
-    id 'org.jetbrains.kotlin.jvm' version '1.3.60'
+    id 'org.jetbrains.kotlin.jvm' version '1.3.61'
```
jeremymailen: This is the right thing to do of course, but it looks like users might bump in kotlin mismatches once our dependency bleeds into their project -- we've encountered things like this before and just set compatibility boundaries in the README.md.

If a project on kotlin 1.3.50 uses this plugin with a `gradle check`, errors like this occur:
```
Caused by: java.io.InvalidClassException: org.jetbrains.kotlin.incremental.IncrementalModuleInfo; local class incompatible: stream classdesc serialVersionUID = 0, local class serialVersionUID = 7715220919552742947
```

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/ExtensionTest.kt
jeremymailen: It's ok. When I make changes to those functionalities I specifically test them.

The integration tests are super valuable, but I still also go through manually testing the plugin on ~ 3 sample projects I have that have different complexities. Gradle plugins are devilish to test and inevitably you run into stuff that's hard to cover synthetically like old gradle/kotlin versions, project structure issues.

jeremymailen: It's ok. When I make changes to those functionalities I specifically test them.

The integration tests are super valuable, but I still also go through manually testing the plugin on ~ 3 sample projects I have that have different complexities. Gradle plugins are devilish to test and inevitably you run into stuff that's hard to cover synthetically like old gradle/kotlin versions, project structure issues.

jeremymailen:
Thanks again for your terrific efforts here! Regarding tests, it's probably wise to have an integration test example both for Android and for generic JVM, but don't go to superhuman efforts to test every environmental possibility -- for example known kotlin version incompatibilities.  It'll be tricky and that in particular will be shifting sand until we make your other redesign suggestion to isolate the class loading. So the right solution in my mind is to engineer robustness as you've suggested instead of test quirks.

mateuszkwiecinski:
> it doesn't seem to pick up the renamed sourceSet. Am I doing something wrong?

@jeremymailen Could you confirm that you wrapped setting sourcesets in a `android` block? So you're configuring AGP's sourcesets, not the uhm.. project/java plugin(?) ones?
I tried to reproduce that behavior and in such setup, if you don't have any files in `src/java/main` folder, the task becames NO_SOURCE so the report doesn't get created.

jeremymailen:
Ok, you are absolutely right. It works when I wrap it in the `android` block.
Also a generic JVM project works as well as far as changing sourceSets.

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/WithGradleTest.kt
```diff
@@ -0,0 +1,50 @@
...
+            val classpath = DefaultClassPath.of(PluginUnderTestMetadataReading.readImplementationClasspath()).asFiles
+            val androidDependencies = listOf(
+                ".*/com\\.android\\..*/.*".toRegex(),
+                ".*/androidx\\..*/.*".toRegex(),
+                ".*/com\\.google\\..*/.*".toRegex()
+            )
+            val noAndroid = classpath.filterNot { dependency -> androidDependencies.any { it.matches(dependency.path) } }
+
+            return defaultRunner(*args)
+                .withPluginClasspath(noAndroid)
```
mateuszkwiecinski: Ok, so here's my suggestion how to simulate the environment without android gradle plugin applied.
I'm removing all android depedencies from the classpath. The code with `dependOn` import fails as expected with 
```
java.lang.NoClassDefFoundError: com/android/build/gradle/internal/tasks/factory/TaskFactoryUtils
```

mateuszkwiecinski:
I forgot to thank you the review and your time. Thanks! 
All issues reported should be already resolved.

> it's probably wise to have an integration test example both for Android and for generic JVM, but don't go to superhuman efforts to test every environmental possibility

yeah, we would probably need to have a separate tests set for Android, JVM and Multiplatform environment. Otherwise all tests share the same classpath and the thing I did doesn't guarantee proper setup completely. I'll try to figure something out later. 
It would be extreamly nice to increase test coverage of the plugin. To have most of the basic functionality covered.

src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
```diff
@@ -20,146 +20,121 @@ import org.jmailen.gradle.kotlinter.tasks.LintTask
...
+
+        // for known kotlin plugins, register tasks by convention.
+        extendablePlugins.forEach { (pluginId, sourceResolver) ->
+            pluginManager.withPlugin(pluginId) {
+
+                val lintKotlin = registerParentLintTask()
+                val formatKotlin = registerParentFormatTask()
+
+                sourceResolver.applyToAll(project) { id, resolveSources ->
+                    val lintKotlinPerVariant = tasks.register("lintKotlin${id.capitalize()}", LintTask::class.java) { lintTask ->
```
jeremymailen: nit: `lintTaskPerSourceSet` would be more accurate -- variant is android specific terminology I believe.

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/WithGradleTest.kt
jeremymailen: I will say this feels a tiny bit fragile. My preferences would be (in a future PR at some point) introduce properly isolated test environments. Possible this would need to be done using two or more test sub-projects which only include kotlinter-gradle via buildscript api as to be more like real projects.

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/ModifiedSourceSetsTest.kt
```diff
@@ -0,0 +1,129 @@
...
+                            id 'kotlin-android'
+                            id 'org.jmailen.kotlinter'
+                        }
+                        
+                        android {
+                            compileSdkVersion 29
+                            defaultConfig {
+                                minSdkVersion 23
+                            }
+                            sourceSets {
```
jeremymailen: Is it possible to test an _added_ sourceSet and see if we pick that up too? I don't know how common it is for people to actually do it, but would be nice for completeness.

mateuszkwiecinski: Huh, I don't know if it is possible to add custom sourceset that would be recognized by AGP 🤔 
If you add that as you would for `java` plugin a _"The SourceSet 'xxxxxx' is not recognized by the Android Gradle Plugin. Perhaps you misspelled something?"_ message pops up. 

Given that, I've added new SourceSet for Kotlin project and for flavor specific SourceSet for Android project.

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/WithGradleTest.kt
mateuszkwiecinski: Initially I thought about separating them by configuration, and there would be separate set of tasks for each environment (i.e. `testAndroid`, `testKotlin`, `estMultiplatform`). So they wouldn't be completely isolated (as they would still share `implementation` configuration), but that would allow us to have control over dependencies that are currently added to `compileOnly`.

However, having tests that introduce completely isolated enviroment sounds even better. I'll try to find some time over the weekend and figure something out.

henrik242:
@mateuszkwiecinski This PR broke `disabledRules` for me (ref. https://github.com/jeremymailen/kotlinter-gradle/issues/130  )

Issue #126, opened 2020-01-19
Use Kotlin DSL?


jmeekhof:
As a kotlin linter plugin, would anyone be opposed to converting the gradle build script for this project to using the kotlin dsl? 

I'm willing to do the work, just want the "blessing" of others involved.

jeremymailen:
Yes, that would be nice. I just hadn't gotten around to it and for a little while the kotlin DSL for gradle was still maturing.

jmeekhof:
Ok, I'll be working on it. I notices that @mateuszkwiecinski has a couple of pull requests that have updated the `build.gradle` file, so I'll base my branches from there so we don't "lose" any updates.

jeremymailen:
Done now, forgot to close!

PR #127, opened 2020-01-20
Convert from groovy to kotlin dsl for build #126

jmeekhof:
This should be applied after #124 has been merged. These changes are applied
following the same versions that were used in that merge.

This PR also uses the kotlin dsl buildSrc object to store all the versions for
all the tooling.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -14,6 +14,15 @@ repositories {
     google()
 }
 
+val pluginId = "org.jmailen.kotlinter"
+val githubUrl ="https://github.com/jeremymailen/kotlinter-gradle"
+val webUrl = "https://github.com/jeremymailen/kotlinter-gradle"
+val projectDescription = "Lint and formatting for Kotlin using ktlint with configuration-free setup on JVM and Android projects"
+
+version = "2.3.0"
+group = "org.jmailen.gradle"
+description = projectDescription
+
 dependencies {
     implementation("com.pinterest.ktlint:ktlint-core:${Versions.ktlint}")
     implementation("com.pinterest.ktlint:ktlint-reporter-checkstyle:${Versions.ktlint}")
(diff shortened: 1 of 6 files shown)


buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,8 @@
+plugins {
+    `kotlin-dsl`
+}
+
+// kotlinDslPluginOptions.experimentalWarning.set(false)
```
jeremymailen: Is it needed or should we remove?

build.gradle.kts
```diff
@@ -0,0 +1,87 @@
...
+val sourcesJar by tasks.registering(Jar::class) {
+    dependsOn(JavaPlugin.CLASSES_TASK_NAME)
+    group = JavaBasePlugin.DOCUMENTATION_GROUP
+    description = "Assembles sources JAR"
+    from(sourceSets.main.get().allSource)
+}
+
+publishing {
+    publications.withType<MavenPublication> {
+        artifact(sourcesJar.get())
```
jeremymailen: Does the artifact `components.java` also get published here by default?

buildSrc/build.gradle.kts
jmeekhof: The comment or the `kotlin-dsl`? The comment isn't needed, and could be removed

build.gradle.kts
jmeekhof: It gets the components via the `sourcesJar` task.

buildSrc/build.gradle.kts
jmeekhof: Without line 5 the following is issued:
```
The `kotlin-dsl` plugin applied to project ':buildSrc' enables experimental Kotlin compiler features.
```

Since we "know" we're using the kotlin dsl, we should probably ignore the warning.

build.gradle.kts
```diff
@@ -0,0 +1,131 @@
...
+    jcenter()
+    google()
+}
+
+val pluginId = "org.jmailen.kotlinter"
+val githubUrl ="https://github.com/jeremymailen/kotlinter-gradle"
+val webUrl = "https://github.com/jeremymailen/kotlinter-gradle"
+val projectDescription = "Lint and formatting for Kotlin using ktlint with configuration-free setup on JVM and Android projects"
+
+version = "2.3.0"
```
jmeekhof: I like using a property for this, as it makes it easier to work locally.

build.gradle.kts
```diff
@@ -0,0 +1,131 @@
...
+    testImplementation("com.nhaarman.mockitokotlin2:mockito-kotlin:${Versions.mockitoKotlin}")
+    testImplementation("org.jetbrains:annotations:${Versions.jetbrainsAnnotations}")
+}
+
+// Required to put the Kotlin plugin on the classpath for the functional test suite
+tasks.withType<PluginUnderTestMetadata>().configureEach {
+    pluginClasspath.from(configurations.compileOnly)
+}
+
+val sourcesJar by tasks.registering(Jar::class) {
```
jmeekhof: Sources jar enables us to publish the sources and javadocs alongside the code.

build.gradle.kts
```diff
@@ -0,0 +1,131 @@
...
+    tags = listOf("kotlin", "ktlint", "lint", "format", "style", "android")
+
+    plugins {
+        named("kotlinterPlugin") {
+            displayName = "Kotlin Lint plugin"
+        }
+    }
+}
+
+artifacts {
```
jmeekhof: This adds the sources and javadocs to the publish task.

build.gradle.kts
```diff
@@ -0,0 +1,131 @@
...
+        pom {
+            name.set(project.name)
+            description.set(project.description)
+            url.set(webUrl)
+
+            scm {
+                url.set(githubUrl)
+            }
+
+            licenses {
```
jmeekhof: I noticed you used an Apache 2.0 license, so I added it to the generated POM

PR #129, opened 2020-01-25
Separate functional tests

mateuszkwiecinski:
In reference to this comment https://github.com/jeremymailen/kotlinter-gradle/pull/124#discussion_r368681815 I've started playing around with the code.

I'm creating PR to get some feedback if that's the separation you had in mind, or the functional tests should be separated even furhter.

Currently the plugin is added to TestKit projects buildscript using existing mechanism involving pluginUnderTestMetadata task, but each project has its own setup. 
I wasn't sure if it is possible to achieve more separated environment using TestKit library 🤔When I've been trying to add completely isolated plugin, but still keep it as a part of the build (for the ease of maintenance and further development) I faced issue where resolved dependency classes were loaded using different classloaders.

FYI: I'll rewrite all scripts to `kts` after #127 is merged.

PR #131, opened 2020-02-08
Fix #130 by using Gradle Property API to configure LintTask and FormatTask

### LINKED ISSUE #130: disabledRules broken in 2.3.0?
After upgrading from 2.2.0 to 2.3.0 `formatKotlin` complains about wildcards: `Format could not fix > [no-wildcard-imports] Wildcard import`, and it tries to order my imports.  This is explicitly disabled in my kotlin dsl config:
```
apply(plugin: "org.jmailen.kotlinter")
kotlinter {
    disabledRules = ["no-wildcard-imports", "import-ordering"]
}
```
Any ideas which change could have triggered this?

mateuszkwiecinski:
Fixes #130

#124 introduced a bug which caused Kotlinter to ignore properties passed via the extension. This PR addresses that issue by declaring all task inputs as Gradle Properties.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.0.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.1.1-all.zip
```
mateuszkwiecinski: I'd like to propose to use distribution version with sources as it is much easier to use Gradle APIs

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/LintTask.kt
```diff
@@ -26,66 +25,38 @@ import org.jmailen.gradle.kotlinter.tasks.lint.LintWorkerRunnable
 @CacheableTask
 open class LintTask @Inject constructor(
     private val workerExecutor: WorkerExecutor
-) : SourceTask() {
+) : ConfigurableKtLintTask() {
 
+    @Optional
     @OutputFiles
-    lateinit var reports: Map<String, File>
+    val reports = mapProperty(default = emptyMap<String, File>())
```
mateuszkwiecinski: Changed behavior: 
Previously, uninitialized reports throwed exception. Now the task simply skips report generation.

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/KotlinProjectTest.kt
```diff
@@ -140,18 +123,6 @@ internal class KotlinProjectTest : WithGradleTest.Kotlin() {
...
-            import org.jmailen.gradle.kotlinter.tasks.LintTask
-
-            task ktLint(type: LintTask) {
-                source files('src')
-                reports = ['plain': file('build/lint-report.txt')]
-                indentSize = 5
-                continuationIndentSize = 7
-                experimentalRules = true
-                disabledRules = ["final-newline"]
-                editorConfigPath = project.rootProject.file(".editorconfig")
```
mateuszkwiecinski: Here are 2 things I'd like to point out:
1. From now on, all tasks treat `.editorconfig` file as one their inputs so the file has to exist. This test started to fail because of that. I decided to move tests related with "Custom Tasks" to separate class and started creating `.editorconfig` file in each test.
2. `editorConfigPath` name suggested it expects a `String` value but in fact, we were passing a File instance.
I changed the task input to use FileProperty, but leaved the name for compatibility reasons.

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/ExtensionTest.kt
```diff
@@ -119,4 +119,34 @@ internal class ExtensionTest : WithGradleTest.Kotlin() {
...
+        projectRoot.resolve("build.gradle") {
+            @Language("groovy")
+            val buildScript = """
+                plugins {
+                    id 'kotlin'
+                    id 'org.jmailen.kotlinter'
+                }
+                
+                tasks.whenTaskAdded {
+                    // configure all tasks eagerly
```
mateuszkwiecinski: Here's test that fails on `2.3.0` that quite clearly shows what causes #130 .

`2.2.0` flow:
1. Apply _Kotlinter_ plugin
2. Register hook (but there are no tasks registered so nothing happens)
3. Configure `kotlinter` extension
4. In project's `afterEvaluate` read from extension and configure all tasks
5. Run registered hook on new tasks

`2.3.0` flow:
1. Apply _Kotlinter_ plugin
2. Register tasks, but not configure them
3. Register hook (triggering registered tasks configuration)
4. Configure tasks by reading from an empty extension
5. Configure `kotlinter` extension

Proposed flow reads from `kotlinter` extension in execution phase so the extension will always be filled in.

build.gradle.kts
```diff
@@ -126,6 +126,9 @@ tasks {
     }
 
     wrapper {
-        gradleVersion = "6.0.1"
+        gradleVersion = "6.1.1"
+        distributionType = Wrapper.DistributionType.ALL
     }
+    register("noOp")
```
jeremymailen: What's this?

src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
```diff
@@ -37,25 +36,37 @@ class KotlinterPlugin : Plugin<Project> {
                 sourceResolver.applyToAll(project) { id, resolveSources ->
                     val lintTaskPerSourceSet = tasks.register("lintKotlin${id.capitalize()}", LintTask::class.java) { lintTask ->
                         lintTask.source(resolveSources)
-                        lintTask.ignoreFailures = kotlinterExtension.ignoreFailures
-                        lintTask.reports = kotlinterExtension.reporters.associate { reporter ->
-                            reporter to reportFile("$id-lint.${reporterFileExtension(reporter)}")
-                        }
-                        lintTask.ktLintParams = kotlinterExtension.toKtLintParams(editorConfigPath())
-                        lintTask.fileBatchSize = kotlinterExtension.fileBatchSize
+                        lintTask.ignoreFailures.set(provider { kotlinterExtension.ignoreFailures })
```
jeremymailen: I assume using this lifecycle user customized values on specific tasks will still take precedence and this documentation is still accurate?
https://github.com/jeremymailen/kotlinter-gradle#customizing-tasks

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/LintTask.kt
jeremymailen: Specifying at least one report output should be mandatory. It uses the report outputs to know if the tasks are up-to-date. Also, no reporting artifacts is probably never what the user wants and they likely got there only by mistake.

.editorconfig of course can remain optional.

build.gradle.kts
mateuszkwiecinski: I mentioned that in the other PR: https://github.com/jeremymailen/kotlinter-gradle/pull/129#discussion_r370923114
When invoking `./gradlew` without arguments it invokes `:help` task by default. This line (and the one below) sets noOp task to be invoked on every empty gradle invocation. 
But yeah, that's just _my preference_ 😛 

🤔 Although,  just realized I should be able to move that to my local Gradle init script instead! So I'll remove that.

src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
mateuszkwiecinski: It should as I had that in mind when writing the code, but I'll add a test for that 👍

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/LintTask.kt
mateuszkwiecinski: >Specifying at least one report output should be mandatory

Thanks for the explanation! So what should happen if user passes an empty map `reports = [:]`?
If I remember well I did such thing in my project because I didn't care about any file outputs. I just wanted the build to fail. 

> It uses the report outputs to know if the tasks are up-to-date

Can you point out the place where those reports are marked as task input?
Plus, as far as I can see LintTask passes its own `HasErrorReporter` so there always would be at least one reporter.

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/CustomTaskTest.kt
```diff
@@ -0,0 +1,188 @@
...
+    @Before
+    fun setUp() {
+        projectRoot = testProjectDir.root.apply {
+            resolve("settings.gradle") { writeText(settingsFile) }
+            resolve("build.gradle") {
+                @Language("groovy")
+                val buildScript = """
+                    plugins {
+                        id 'kotlin'
+                        id 'org.jmailen.kotlinter'
```
mateuszkwiecinski: Here's one more line I would like to discuss. When I was performing code changes initially I wanted to read values from the extension, but then I realized that one of plugin capabilities is to create standalone tasks: https://github.com/jeremymailen/kotlinter-gradle#custom-tasks
so the plugin itself doesn't have to be applied. Kotlinter dependency just needs to be added to project conifuration.

```suggestion
```

Therefore I started wondering if I should remove the plugin application from tests.

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/CustomTaskTest.kt
```diff
@@ -0,0 +1,198 @@
...
+                import org.jmailen.gradle.kotlinter.tasks.LintTask
+    
+                task reportsNotConfigured(type: LintTask) {
+                    source files('src')
+                }
+                
+                task reportsEmpty(type: LintTask) {
+                    source files('src')
+                    reports = [:]
+                }
```
mateuszkwiecinski: FYI: I added test covering current behavior. I'll update them once we agree what should be a valid behavior for an empty map.

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/ExtensionTest.kt
```diff
@@ -119,4 +119,66 @@ internal class ExtensionTest : WithGradleTest.Kotlin() {
...
+        projectRoot.resolve("build.gradle") {
+            @Language("groovy")
+            val script = """
+                kotlinter {
+                    disabledRules = ['filename']
+                }
+                
+                lintKotlinMain {
+                    disabledRules = ['final-newline']
+                }
```
mateuszkwiecinski: https://github.com/jeremymailen/kotlinter-gradle/pull/131/files#r376768433

✅ Done
I've added simple case for `disabledRules` only, but if needed I can add more detailed checks for every property, if needed.

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/LintTask.kt
jeremymailen: `[:]` would also be an error I think. In your project, is your `lintKotlin` task ever "UP-TO-DATE"?

`reports` is annotated as `@OutputFiles` which means in factors in to incremental build. You should see this in a normal project with at least one report.
1. `gradle -i lintKotlin` from a clean state and notice it runs the task
2. `gradle -i lintKotlin` again and notice the task is up to date and does not rerun
3. delete the report output files and repeat the process noting the task reruns

https://docs.gradle.org/current/userguide/more_about_tasks.html

mateuszkwiecinski: > In your project, is your lintKotlin task ever "UP-TO-DATE"?

I confirm all tasks normally became `UP-TO-DATE` 🤷‍♂ I've also commited new tests confirming that behavior. 

> You should see this in a normal project with at least one report.

I confirm that's what happens with an empty reports map. (Except the `delete the report output files` part as I cannot delet any files because they don't exist ofc 😛)

I think you're referring to the behavior change introduced in: https://github.com/gradle/gradle/pull/9896 but isn't applicable here. The `LintTask`, both in previous version as the one I'm suggesting declares Outputs and in both cases they both properly handles empty map case (so the output is declared, it exists, but it is empty so it never changes) therefore Gradle takes only task inputs into account when running incremental build. 
Documentation seems to confirm that behavior:
> As part of incremental build, Gradle tests whether any of the task inputs or outputs **has** **changed** since the last build. If they haven’t, Gradle can consider the task up to date and therefore skip executing its actions. Also note that incremental build won’t work unless a task has **at least one task output**, although tasks usually have at least one input as well.

and I would name that consistent with the fact that `ListProperty` and `MapProperty` always have default values where other properties throw exception when accessed uninitialized (unless marked as `@Optional`.), having an empty collection is considered a valid case.
Or at least that's my understanding of how `Inputs` and `Outputs` work 🤷‍♂ 

Therefore, if the reports property was required only for the sake of incremental build, and empty map has been treated as a valid value I'd like to propose to keep that behavior and unify it with the _uninitialized_ value case. That would be somehow backward compatbile, right? Every developer that had an empty map passed would still have their code working. There are no developers who have reports set to `null` as that previously resulted in an exception.

jeremymailen: If an empty value doesn't interfere with incremental build and some developers find it preferable, I'm good with keeping it.

Generally I think output artifacts are a good idea for almost any task -- a build step produces _something_ after all. Many people are using these reports to display the build results in CI.

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/CustomTaskTest.kt
jeremymailen: I wouldn't worry too much about that case.

henrik242:
@mateuszkwiecinski Thank you so much for this! :)

PR #133, opened 2020-02-18
Fix #132 by respecting indent_size from .editorconfig

### LINKED ISSUE #132: Configuring indent_size from .editorconfig broken
No longer honors the value specified, although configuring it in the `kotlinter` extension still works.

This functionality appears to have broken between 2.0.0 and 2.1.0, so somewhere in #111 things went wrong. Needs investigation. Could have been the ktlint upgrade or changes to this plugin.

mateuszkwiecinski:
Fixes #132 

The reson why properties set in `.editorconfig` weren't respected is that we were always passing them in ktlint's userData which are higher in hierarchy than those set in `.editorconfig`.

In addition I removed support for `continuation_indent_size` as it is no longer recognized by ktlint https://github.com/pinterest/ktlint/issues/171 (I could mark it as deprecated though 🤔WDYT?)

### DIFF
--- a/src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterExtension.kt
+++ b/src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterExtension.kt
@@ -5,8 +5,6 @@ import org.jmailen.gradle.kotlinter.support.ReporterType
 open class KotlinterExtension {
     companion object {
         const val DEFAULT_IGNORE_FAILURES = false
-        const val DEFAULT_INDENT_SIZE = 4
-        const val DEFAULT_CONTINUATION_INDENT_SIZE = 4
         val DEFAULT_REPORTER = ReporterType.checkstyle.name
         const val DEFAULT_EXPERIMENTAL_RULES = false
         val DEFAULT_DISABLED_RULES = emptyArray<String>()
@@ -16,9 +14,7 @@ open class KotlinterExtension {
     /** Don't fail build on lint issues */
     var ignoreFailures = DEFAULT_IGNORE_FAILURES
 
-    var indentSize = DEFAULT_INDENT_SIZE
-
-    var continuationIndentSize = DEFAULT_CONTINUATION_INDENT_SIZE
+    var indentSize: Int? = null
 
     var reporters = arrayOf(DEFAULT_REPORTER)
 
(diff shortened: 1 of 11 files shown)


src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterExtension.kt
```diff
@@ -16,9 +14,7 @@ open class KotlinterExtension {
     /** Don't fail build on lint issues */
     var ignoreFailures = DEFAULT_IGNORE_FAILURES
 
-    var indentSize = DEFAULT_INDENT_SIZE
-
-    var continuationIndentSize = DEFAULT_CONTINUATION_INDENT_SIZE
```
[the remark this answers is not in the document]
jeremymailen: If you wish to remove it from the readme, you can or I'll make those updates when I push the version bump.

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/EditorConfigTest.kt
```diff
@@ -0,0 +1,93 @@
...
+                }
+                
+            """.trimIndent()
+                writeText(buildScript)
+            }
+        }
+    }
+
+    @Test
+    fun `lintTask uses default indentation if editorconfig absent`() {
```
jeremymailen: oh good, so that's ktlint default when there isn't userData or an .editorconfig.

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/EditorConfigTest.kt
```diff
@@ -0,0 +1,93 @@
...
+            )
+        }
+
+        build("lintKotlin").apply {
+            assertEquals(TaskOutcome.SUCCESS, task(":lintKotlinMain")?.outcome)
+        }
+    }
+
+    @Test
+    fun `plugin respects indentSize set in editorconfig`() {
```
jeremymailen: What about the test case that when both kotlinter.indentSize is specified and there's an .editorconfig with the param, the extension wins? Based on the code I expect this to work of course.

src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterExtension.kt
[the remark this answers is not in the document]
mateuszkwiecinski: Done ✅ 
Added deprecation messages, both for tasks and for the extension. Removed all `continuationIndentSize` mentions from Readme.

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/EditorConfigTest.kt
mateuszkwiecinski: Done ✅ Added precedence checks for both `disabledRules` and `indentSize`

PR #137, opened 2020-05-18
Pre push hook

dinomite:
This adds a task that will install a Git pre-push hook which runs the `lintKotlin` task and, if that has a non-zero exit, the `formatKotlin` task.


dinomite:
I'm getting started on a functional test, but wanted to make sure this was desired and seemed like a good approach.

jeremymailen:
Thank you for your contribution!

I'm open to including git hook install if a lot of people would like to have this functionality. In the past I've stayed away from things outside the core lint and format functionality in keeping with the simple aesthetic of ktlint. I don't know if there's general agreement on one right workflow. Should it quietly format on pre-commit or even every save a la `prettier`? Should it just lint at different phases of the workflow to prompt you to take action instead of wasting time on a failed build?

I live that you made efforts not to overwrite existing hook commands the user might have.

Let me know your thoughts and if you've heard feedback from other users of the plugin. Thanks again.

dinomite:
Thanks for the input—I'm certainly ok with waiting on broader community support before moving ahead.

My org standardized on `kotlinter-gradle` after finding it to be far faster than `ktlint-gradle`.  We use a pre-commit hook like the one herein, which runs `lintKotlin` and, if that exist non-zero, runs `formatKotlin`, but leaves you to commit those changes.  I wonder whether just running `formatKotlin` and continuing with the commit might be better to your point of not wasting time on failed builds.  Another option would be to put that into a pre-push hook that either auto-commits any changes or fails the push leaving the user to commit the results of the `formatKotlin`.

jeremymailen:
Ok, thank you for the feedback @dinomite and good to hear the details about how your team is using the plugin and workflow. Let's move ahead with the feature!

I think there's a reasonable case to be made for what you've implemented here and how your team uses it:
1. pre-push is a final backstop that's done the least frequently of the places we could insert this check (less overhead for the average user)
2. it avoids auto-committing anything. I've seen bugs filed elsewhere about pre-commit hooks indiscriminately adding changes to the commit the user didn't intend, sometimes ones the hook didn't actually make. Also, `formatKotlin` isn't capable of fixing all lint errors, so "auto-fix" would not be complete anyhow.

The only possible wrinkle is if the `formatKotlin` call on failure mixes new changes into any dirty or staged changes the user might already have. I think the console output makes it clear what was changed though and they would be changes the user would have to make anyway to successfully creating a passing PR. It's ok with me if this makes sense to you as well.

I'll give it a review and test it out.

build.gradle.kts
```diff
@@ -19,7 +19,7 @@ val githubUrl ="https://github.com/jeremymailen/kotlinter-gradle"
 val webUrl = "https://github.com/jeremymailen/kotlinter-gradle"
 val projectDescription = "Lint and formatting for Kotlin using ktlint with configuration-free setup on JVM and Android projects"
 
-version = "2.3.2"
+version = "2.3.3-SNAPSHOT"
```
jeremymailen, line 22: No need to change this, I'll update the version when I update the readme and push out the new release.

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/InstallPrePushHookTask.kt
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Install or update the kotlinter-gradle pre-push hook.
+ */
+open class InstallPrePushHookTask : DefaultTask() {
+    @TaskAction
+    fun run() {
+        val dotGitDir = findGitDir(project.rootDir)
+        logger.info(".git directory: $dotGitDir")
+        val gradlew = findGradlew(project.rootDir).path
```
jeremymailen, line 17: I know it's crazy not to have a gradle wrapper installed, but what if we just abstracted this to `gradleCommand` and resolve it to `gradlew` or `gradlew.bat` when available, otherwise `gradle`? Do remember the extension on Windows!

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/InstallPrePushHookTask.kt
```diff
@@ -0,0 +1,91 @@
...
+import java.io.File
+
+/**
+ * Install or update the kotlinter-gradle pre-push hook.
+ */
+open class InstallPrePushHookTask : DefaultTask() {
+    @TaskAction
+    fun run() {
+        val dotGitDir = findGitDir(project.rootDir)
+        logger.info(".git directory: $dotGitDir")
```
jeremymailen, line 16: For simplicity to the user, probably sufficient at the end to info level output either `created .git/hooks/pre-push` or `updated .git/hooks/pre-push`

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/InstallPrePushHookTask.kt
```diff
@@ -0,0 +1,91 @@
...
+            logger.info("Writing hook to empty file")
+            prePushHookFile.writeText(generateHook(gradlew, addShebang = true))
+        } else {
+            val prePushHookFileContent = prePushHookFile.readText()
+            val startIndex = prePushHookFileContent.indexOf(startHook)
+            if (startIndex == -1) {
+                logger.info("Appending hook to end of existing non-empty file")
+                prePushHookFile.appendText(generateHook(gradlew))
+            } else {
+                logger.info("Replacing existing hook")
```
jeremymailen, line 45: This is the case where we replace our own hook snippet? Is the use case I newer version of kotlinter-gradle with revised hook logic?

src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
```diff
@@ -11,9 +11,12 @@ import org.gradle.api.plugins.JavaPluginConvention
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.SourceSet
 import org.gradle.api.tasks.SourceSetContainer
+import org.gradle.language.base.plugins.LifecycleBasePlugin
```
jeremymailen, line 14: Unused import?

jeremymailen:
A couple small lints
```
> Task :lintKotlinMain FAILED
Lint error > src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/InstallPrePushHookTask.kt:3:1: [import-ordering] Imports must be ordered in lexicographic order without any empty lines in-between
Lint error > src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt:14:1: [no-unused-imports] Unused import
Lint error > src/main/kotlin/org/jmailen/gradle/kotlinter/support/FileFinders.kt:3:1: [import-ordering] Imports must be ordered in lexicographic order without any empty lines in-between
```
Somehow it's only showing in travis-ci and not in the checks here. The irony does not escape me 😂 .

jeremymailen:
Ok, I guess Travis needed my OAuth token to be refreshed, checks are back on the menu.

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/InstallPrePushHookTask.kt
```diff
@@ -0,0 +1,91 @@
...
+         * Generate the hook script
+         */
+        private fun generateHook(
+            gradlew: String,
+            addShebang: Boolean = false,
+            includeEndHook: Boolean = true
+        ): String {
+            return """
+                    ${if (addShebang) shebang else ""}
+                    $startHook
```
jeremymailen, line 83: You might need to `trimIndent()` or `trimMargin()` for these literals. Hook formatting ends up looking a little weird:
```sh
        ##### KOTLINTER HOOK START #####
        GRADLEW=/user/myrepo/gradlew

        
$GRADLEW lintKotlin

status=$?
if [ "$status" != 0 ] ; then
    echo 1>&2 "\nlintKotlin found problems; running formatKotlin command..."
    $GRADLEW formatKotlin
    exit 1
fi

        ##### KOTLINTER HOOK END #####
```

dinomite:
@jeremymailen I'll get back to address your comments soon!

@shabtaisharon Sure, I think that's a fair argument.  The argument for a pre-push is that it prevents the linter's auto-format for incidentally breaking things (see [@jeremymailen's comment](https://github.com/jeremymailen/kotlinter-gradle/pull/137#issuecomment-635117145))

jeremymailen:
Yes @shabtaisharon I think that's the tricky part with including workflow hooks. Different people will do it different and valid ways. Some even use a file watcher to format a file as they save it. I could see pre-commit being an obstacle for some people who want to be able to quickly commit some code so they can switch branches.

Perhaps it's fair to include a couple canonical workflows?
- Pre-push for orgs that like a backstop
- Pre-commit for orgs that want a fire and forget format on commit

Do those sound like two reasonable approaches? Guaranteed we'll forget to account for some people's needs, but we will get ongoing feedback and be able to adjust.

shabtaisharon:
Yes, i like your idea @jeremymailen Thank you!

dinomite:
Yep, I think providing both is a good way to go.  I'll get that done along with the other updates soon.

src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
```diff
@@ -89,6 +93,12 @@ class KotlinterPlugin : Plugin<Project> {
             it.group = "formatting"
             it.description = "Formats the Kotlin source files."
         }
+
+    private fun Project.registerPrePushHookTask() =
+        tasks.register("installKtLintPrePushHook", InstallPrePushHookTask::class.java) {
+            it.group = VERIFICATION_GROUP
```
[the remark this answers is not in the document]
dinomite: Good call 👍

src/main/kotlin/org/jmailen/gradle/kotlinter/support/FileFinders.kt
```diff
@@ -0,0 +1,40 @@
+package org.jmailen.gradle.kotlinter.support
+
+import com.google.common.annotations.VisibleForTesting
+import org.gradle.api.GradleException
+import java.io.File
+
+/**
+ * Find the nearest .git directory
+ */
+fun findGitDir(dir: File): File {
```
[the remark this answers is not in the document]
dinomite: Yeah, I think reasonable to only look in the project root for this & gradlew

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/InstallPrePushHookTask.kt
dinomite: How about making the final message a `quiet` and these earlier ones `info`?

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/InstallPrePushHookTask.kt
dinomite: Ooh, great point about `…bat`.  I'll write something up that looks for `gradlew(.bat)?` and, if there isn't one, attempts to use a system-provided `gradle(.bat)?`.  I wish there were a way to just have Gradle tell you where it's executing from…

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/InstallPrePushHookTask.kt
dinomite: Exactly.

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/InstallPrePushHookTask.kt
dinomite: Actually, I don't know how to make that work—we don't have a shell with which to find `gradle`, the `PATH` environment variable is around, so we could manually search that, but then I don't have a clue what to do on Windows.  Suggestions?

dinomite:
Ok, minor updates handled, one open question re: finding system Gradle on.

I'll get to providing two tasks, one for pre-commit and one for pre-push, soon.

dinomite:
Also, I cannot at all figure out how to make `.trimIndent()` or `.trimMargin()` work properly.  Open to suggestions there…

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/InstallPrePushHookTask.kt
jeremymailen: Two possibilities.
1. We just do `gradle` and assume it's in their path, which we hope it is if they don't have wrapper installed.
2. We simply require gradle wrapper. If we don't find it, we exit with an error saying so.

Open to either. Which do you like?

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/InstallPrePushHookTask.kt
jeremymailen: The user is only likely to see `quiet`. If you think there's useful troubleshooting info you'd ask for if someone is having trouble, I'd include it in `debug` level. We do that elsewhere in the plugin and it did help in a couple of issues filed.

src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
```diff
@@ -37,6 +38,7 @@ class KotlinterPlugin : Plugin<Project> {
 
                 val lintKotlin = registerParentLintTask()
                 val formatKotlin = registerParentFormatTask()
+                registerPrePushHookTask()
```
jeremymailen, line 41: You'll need to guard this with
```kotlin
if (project.rootProject == project) {
    registerPrePushHookTask()
}
```
Otherwise in multi-module builds the task will be registered on each module and running it from the top triggers multiple executions.

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/InstallPrePushHookTask.kt
```diff
@@ -0,0 +1,114 @@
...
+        } else {
+            val prePushHookFileContent = prePushHookFile.readText()
+            val startIndex = prePushHookFileContent.indexOf(startHook)
+            if (startIndex == -1) {
+                logger.info("Appending hook to end of existing non-empty file")
+                prePushHookFile.appendText(generateHook(gradleCommand))
+            } else {
+                logger.info("Updating existing kotlinter-installed hook")
+                val endIndex = prePushHookFileContent.indexOf(endHook)
+                val newPrePushHookFileContent = prePushHookFileContent.replaceRange(
```
jeremymailen, line 48: This appears to bump it down one line each time it gets replaced.

src/test/kotlin/org/jmailen/gradle/kotlinter/tasks/InstallPrePushHookTaskTest.kt
```diff
@@ -0,0 +1,56 @@
+package org.jmailen.gradle.kotlinter.tasks
+
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Test
+
+
```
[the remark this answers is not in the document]
dinomite: Ugh, I can't wait until the Kotlin plugin brings IJ's orderings inline with ktlint.

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/PrePushHookTest.kt
```diff
@@ -0,0 +1,108 @@
...
+import org.gradle.testkit.runner.TaskOutcome.SUCCESS
+import org.jmailen.gradle.kotlinter.functional.utils.resolve
+import org.jmailen.gradle.kotlinter.functional.utils.settingsFile
+import org.jmailen.gradle.kotlinter.tasks.InstallPrePushHookTask
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+import java.io.File
```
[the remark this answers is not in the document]
dinomite: Ugh, I can't wait until the Kotlin plugin brings IJ's orderings inline with ktlint.

src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
dinomite: Great catch

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/InstallPrePushHookTask.kt
dinomite: Oh, the first is a fine idea—I forgot that we don't actually need to interact with Gradle herein, just the script we're writing needs it at invocation time.  Easy, if the wrapper isn't found, I'll just use `gradle`.

dinomite:
I think this is complete.  I added a small section to the readme describing the Git hooks.

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/InstallPrePushHookTask.kt
```diff
@@ -23,38 +26,38 @@ open class InstallPrePushHookTask : DefaultTask() {
                 mkdir()
             }
         }
-        logger.info("hookDir: $hookDir")
+        logger.quiet("hookDir: $hookDir")
```
jeremymailen, line 29: you might have meant for this to be `info` level?

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/InstallPrePushHookTask.kt
```diff
@@ -23,38 +26,38 @@ open class InstallPrePushHookTask : DefaultTask() {
...
-                val endIndex = prePushHookFileContent.indexOf(endHook)
-                val newPrePushHookFileContent = prePushHookFileContent.replaceRange(
+                val endIndex = hookFileContent.indexOf(endHook)
+                val newHookFileContent = hookFileContent.replaceRange(
                     startIndex,
                     endIndex,
                     generateHook(gradleCommand, includeEndHook = false)
                 )
-                prePushHookFile.writeText(newPrePushHookFileContent)
+                hookFile.writeText(newHookFileContent)
```
jeremymailen, line 56: Should the two different hooks be different content? Intuitively I might expect a pre-commit hook to simply `formatKotlin` and continue the commit, aborting only if auto-formatting is not possible.
The rationale is a pre-push hook is a backstop to prevent wasting a build, but you chose to have your workflow otherwise unaltered. A pre-commit hook feels more like a "just quietly take care of it" situation. What do you think?

dinomite: That's a reasonable distinction to me, updates shortly

dinomite:
Split out two separate hooks—pre-commit just runs format, pre-push lints and, if formatting is necessary, does so & notifies the user that they need to commit the result

dinomite:
Checking on those build errors; I've still got more to learn about plugins.

dinomite:
I'll try to take a look at this next week.

How about `git diff --name-only --cached` to get the files to process, run `formatKotlin` on those, then `git add` that same set?

jeremymailen:
The commands `ktlint` hooks rely on are as you indicate
pre-commit: `git diff --name-only --cached --relative` filtering to only *.kt* files)
pre-push: `git diff --name-only HEAD origin/$(git rev-parse --abbrev-ref HEAD)`

For a moment I got to thinking maybe recommending people use the native `ktlint` hooks would be a good path, but I do think that because kotlinter-gradle is integrated at the code level and we can react to each individual file, it's possible to implement better functionality to automatically add each corrected file to the commit.

So yeah, I think we have the opportunity to do something nice here for the community at the cost of the slight messiness of having to interact with git commands. I wonder if we created a new type of format task which took it's input list from a shell command and ran a second shell command on the list of fixed files (if any), failing if un-fixable files were encountered. There would be some complexity if multiple shell commands are needed in pre or post, but maybe there's a good library for running shell scripts via Kotlin?

PR #139, opened 2020-06-03
update ktlint (0.37.0)

olegivo:
- update ktlint (0.37.0)
- kotlin (1.3.70)


jeremymailen:
Thank you for the assist. I'm inclined to see if there's a forthcoming minor version to ktlint for fixing this: https://github.com/pinterest/ktlint/issues/682

olegivo:
But now I know how the gradle plugin tests work.  
When the test fell, I could not figure out which side the error was, but then I ran the same test for pure ktlint and realized that the problem was in it

jeremymailen:
Superceded by https://github.com/jeremymailen/kotlinter-gradle/pull/141

PR #140, opened 2020-06-10
Temporarily disable pre-commit hook

jeremymailen:
To be helpful for automatically formatting, it would need to add the changes to the commi, which isn't a style of format command we have yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,8 +113,7 @@ Granular tasks exist for each source set in the project: `formatKotlin`*`SourceS
 
 ### Git Hooks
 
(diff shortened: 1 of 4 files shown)


SleipRecx:
Will this ever be enabled again?

jeremymailen:
I think it will need a new type of task to implement it, see https://github.com/jeremymailen/kotlinter-gradle/issues/155.

I haven't had much time to tackle it lately, but I'm open to pull requests.
Although it seems like it would be a simple thing, I think it's going to require careful testing to make sure it has no unexpected bugs in different people's environments.

PR #147, opened 2020-07-01
Only write to the hook file if necessary

dinomite:


jeremymailen:
Should we customize the `updateToDate` so that the task reports as not being run? I realize it's somewhat superficial but if a frequent task such as `check` is dependent, it might give a clearer build result output.

Maybe it would be a quick check if the file had `##### KOTLINTER <version number> HOOK START #####`?

Thank you for your patience. Normally I try to get feedback to contributors same week, but a lot going on last couple of weeks!

dinomite:
Good idea on actually complying with Gradle's expectations re up-to-date checks.

Version in the hook is a good idea—we could either use the plugin version if there's a way to get that, or just have a manually-incremented version in `InstallHookTask`.  Thoughts?

jeremymailen:
Yeah I think the plugin version works fine. It'll get rewritten each revision whether it needs it or not, but that should be ok.

dinomite:
Do you have ideas for how to get the version from within the plugin code itself?

jeremymailen:
I don't know if there is an easy programmatic way, reference: https://stackoverflow.com/questions/41689977/how-can-i-determine-the-android-gradle-plugin-version-in-use-from-within-my-own

I'm fine with introducing a const into the plugin code and keeping it in sync with the build plugin version.

dinomite:
@jeremymailen Ended up writing the version from the `build.gradle.kts` to a resource file at build time and reading that in the hook task code.  Bit of a change, split diff will be clearest.

jeremymailen:
Thank you! Made some adjustments and incorporated: https://github.com/jeremymailen/kotlinter-gradle/pull/158
Will push a release.

PR #153, opened 2020-08-19
Remove configuration of .editorconfig in kotlinter

### LINKED ISSUE #151: KotlinPlugin should not enforce .editorconfig of the rootProject to be passed to Ktlint
**Current:** Creating a `.editorconfig` at the root of the project with a section name matching a directory name will never match.

Example of project structure: 

```
.editorconfig
build.gradle.kts
src/main/FileWithLintError.kt
src/test/NonToBeChecked.kt
```

Example of `.editorconfig`:
```
root = true

[**/main/**.{kt,kts}]
max_line_length=5
```

Ktlint will not apply the parameter from the `.editorconfig`. The purpose here is to have a specific parameter for main sources and another one for test sources.

**Expected:**:  Creating a `.editorconfig` at the root of the project with a section name matching a directory name will match and apply the parameter from the section.

The issue is that the the `.editorconfig` from the rootProject will be passed as editorconfig path if one exist
- https://github.com/jeremymailen/kotlinter-gradle/blob/master/src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt#L163
- https://github.com/jeremymailen/kotlinter-gradle/blob/master/src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt#L62

fast forward, this parameter is passed to ktlint in the LintWorkerRunnable : 
- https://github.com/jeremymailen/kotlinter-gradle/blob/master/src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/lint/LintWorkerRunnable.kt#L71

fast forward, the editor path will be used as `alternativeEditorConfig` in ktlint
- https://github.com/pinterest/ktlint/blob/master/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/internal/EditorConfigLoader.kt#L59

The file path (the file that is being check) will be change, as `alternativeEditorConfig` is not null: 
instead of being `src/main/FileWithLintError.kt` it will be `<fullpath of .editorconfig>/FileWithLintError.kt`.
This new path is used only for section matching, and in this case will not match again the section from our .editorconfigfile.
as `<fullpath of .editorconfig>/FileWithLintError.kt`. will not match against `[**/main/**.{kt,kts}]`

You can confirm this behaviour (path change of the file analysed) using the test of ktlint: https://github.com/pinterest/ktlint/blob/master/ktlint-core/src/test/kotlin/com/pinterest/ktlint/core/internal/EditorConfigLoaderTest.kt#L267

**Proposed workaround:** Creating a .editorconfig in all directories excepting in the root directory. It will work only with project with with submodules.

**Proposed fixed:** 
The editorconfig path should [not be enforced by default but being a parameter of the plugin extension](https://github.com/jeremymailen/kotlinter-gradle/blob/master/src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt#L62,75) with null value by default (ktlint should find the root editorconfig file anyway).

jeremymailen:
Fixes #151

Ktlint will already pick up .editorconfig files in the proper directories.
Us configuring it actually causes path bugs.


jeremymailen:
Although one wrinkle worth mentioning is when .editorconfig is not configured as a build input to gradle, changes to it are not picked up by incrementals builds. For example a `gradle clean lintKotlin` may be required after changing .editorconfig.

Issue #155, opened 2020-08-26
Make pre-commit hook only look at changed files


dinomite:
kotlinter-gradle has most of the wiring necessary for a pre-commit hook, but [it was disabled](https://github.com/jeremymailen/kotlinter-gradle/pull/137/#pullrequestreview-427739206) because it checks _all_ files, not just those that are part of the to-be-committed changeset.

Make the pre-commit hook only examine files that are part of the changeset and wire it as a new task.

jeremymailen:
Yeah, it's a good idea. Been meaning to get to this, but after all the work to get kotlin 1.4 compatibility out I took a bit of a hiatus.

Noctiflore:
I saw you plan to do it using a git diff command to get the list of changed files, but modifying files and staging them without staging unwanted modifications from the working directory would still be hard.
Why don't stash the workdir (and unstash before real commit), as done [here](https://medium.com/@alistair.cerio/android-ktlint-and-pre-commit-git-hook-5dd606e230a9) ? git diff can still be used to check if the hook needs to be run at all, and if there are unstaged modifications to stash.

jeremymailen:
Thank you @Noctiflore, these are good references. Will definitely try to make use of them.

Noctiflore:
Some other useful readings and details:
- [my first link](https://medium.com/@alistair.cerio/android-ktlint-and-pre-commit-git-hook-5dd606e230a9) is a good starting point but fails with partially staged files.
- [git-format-staged](https://github.com/hallettj/git-format-staged) discusses the different alternatives ; a bit outdated, but still useful.
- [one of the answers here](https://stackoverflow.com/questions/8506334/what-to-stage-after-a-pre-commit-hook-was-run) propose to use git merge -Xtheirs instead of git stash pop, which may not be what is expected (overwrite linter changes)
- [this article](https://www.darrenlester.com/blog/git-pre-commit-stash) propose a good solution to handle a stash, and the specific case when stash push is a no-op.

I ended up writing this, inspired from these articles and git documentation:
```shell
#!/bin/bash

# Git pre-commit hook applying Ktlint on staged files
# (using user-defined gradle task)

STASH_NAME=pre-commit-$(date +%s)-$RANDOM
if ! git diff-files --quiet ; then
    # dirty workdir: stash to work on staged files only
    git stash push -q --keep-index -m "$STASH_NAME"
fi

echo "Pre-commit hook: Running ktlint (fix)"
[...]

PR #172, opened 2021-03-06
fix: Pre-Commit hook task not being registered.

Querschlag:
There is a function to install the plugin's pre-commit hook task, but it is never used.
This results in `Task with path 'installKotlinterPreCommitHook' not found in root project` when using `dependsOn("installKotlinterPreCommitHook")` in a project.

### DIFF
--- a/src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
+++ b/src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
@@ -24,6 +24,7 @@ class KotlinterPlugin : Plugin<Project> {
 
         if (this == rootProject) {
             registerPrePushHookTask()
+            registerPreCommitHookTask()
         }
 
         // for known kotlin plugins, register tasks by convention.


jeremymailen:
Thank you for the PR @Querschlag. You are quite right that this task hasn't been registered.

The reason is: we weren't sure it's the right functionality for pre-commit. It will blindly format all files in the project, potentially changing files that weren't involved in the commit. We had a todo item to improve the hook such that it would only format the files that were staged for the commit. This of course would require making a new kind of format task which takes a list of files. (instead of targeting a sourceSet) and does not write a report file.

If we feel like the native "format everything" approach is still ok, we could go ahead and register it. We would also want to document it here: https://github.com/jeremymailen/kotlinter-gradle#git-hooks

Open to people weighing in with their thoughts here.

Mahoney:
The relevant issue is this one: https://github.com/jeremymailen/kotlinter-gradle/issues/155

jeremymailen:
Will close for now since it requires more work to complete the implementation.

PR #177, opened 2021-03-27
Allow capturing the status of the commands

Mahoney:
`shebang` includes a `set -e` which means neither of the `status=$?` will ever happen if `$? > 0`.

`set -e` is good practise; simplest fix is to get rid of the explicit status checking and instead run the command in an `if ! $CMD ; then`


Mahoney:
I'd love to have written a proper test for this (i.e. one that actually commits some badly formatted code, pushes it and proves that when `lintKotlin` fails on pre-push then `formatKotlin` is run) but I just don't have time to do it... wouldn't be that hard though, you can create a gradle project with the plugin in a dir in /tmp, check it out to another dir in /tmp and try and push it...

jeremymailen (approved):
Thank you @Mahoney, you're a shell wizard!

Understood on the test. Things with a lot of system and external dependencies do get intricate to test in a realistic way and often, given the modest cycle time for kotlinter, I opt for testing it in a real environment.

PR #181, opened 2021-04-11
Bump gradle to 7.0

Goooler:


jeremymailen (commented):
Thank you -- if you wouldn't mind, could you also update the `wrapper` spec in the build.gradle.kts?

Goooler:
> Thank you -- if you wouldn't mind, could you also update the `wrapper` spec in the build.gradle.kts?

Done.

Issue #173, opened 2021-03-15
JCenter/Bintray are going down next month. Looks like this module is hosted there. Find a new home?


gefilte:
https://www.infoq.com/news/2021/02/jfrog-jcenter-bintray-closure/

So by end of April this hosting will be unavailable. Have a new home in mind?

jeremymailen:
Thanks for the heads up!

We don't publish to to jcenter (gradle plugins portal is where the artifacts get pushed).

It looks like we can switch over to pulling all dependencies from `mavenCentral()` with the noteworth exception of trove4j https://youtrack.jetbrains.com/issue/KT-44730. Assume that will get resolved in time.

I'll make the switch in a `3.4.0` release. We've got a bit of time since it looks like jcenter will continue to serve packages until Feb 1, 2022.

jeremymailen:
3.4.0 didn't fully get rid of jcenter. It's still providing trove4j... just waiting on the owners to resolve, but we've got time.

jeremymailen:
Fixed in #184

PR #184, opened 2021-04-15
Bump versions

Goooler:


Goooler:
I found two test failures run with Windows, because the regex is not compatible with Windows path

image

Goooler:
Regex is here, paths is splited with `\` in Windows instead of `/`, what you declared should be `"((/|\\\\).*\\.kt):\\d+:\\d+"`

image

Goooler:
When I ran this test again after corrected the `pathPattern`, new case occured, `filePath` miss `C:\` prefix, the file of it could not be specified, the `File(filePath).exists()` is false. 
You may need to test in Windows and fix it.

image

build.gradle.kts
```diff
@@ -12,7 +12,6 @@ plugins {
 repositories {
     mavenCentral()
     google()
-    jcenter() // remove after trove4j resolved https://youtrack.jetbrains.com/issue/KT-44730
```
jeremymailen, line 15: This is still needed, no?

jeremymailen:
True, I've not tried to execute the tests on windows. I will file a bug.

build.gradle.kts
Goooler: Yeah, need not now.

jeremymailen: Oh good, they finally got it moved
https://youtrack.jetbrains.com/issue/IDEA-261387

jeremymailen (approved):
ty

jeremymailen:
Fixes #173

PR #185, opened 2021-04-15
Simplify gradle config

Goooler:
- Use extension method to simplify `withType`
- `apiVersion` & `languageVersion` need not be declared, they will follow the kotlin plugin' version by default.
- `wrapper` can be removed also as below.

image

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -65,21 +65,15 @@ tasks {
         dependsOn(generateVersionProperties)
     }
 
-    withType<KotlinCompile>().configureEach {
+    withType<KotlinCompile> {
         kotlinOptions {
-            apiVersion = "1.4"
-            languageVersion = "1.4"
             jvmTarget = JavaVersion.VERSION_1_8.toString()
         }
     }
 
     // Required to put the Kotlin plugin on the classpath for the functional test suite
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -65,22 +65,16 @@ tasks {
...
         kotlinOptions {
-            apiVersion = "1.4"
-            languageVersion = "1.4"
             jvmTarget = JavaVersion.VERSION_1_8.toString()
         }
     }
 
     // Required to put the Kotlin plugin on the classpath for the functional test suite
-    withType<PluginUnderTestMetadata>().configureEach {
+    withType<PluginUnderTestMetadata> {
```
mateuszkwiecinski, line 75: Wouldn't `withType` configure tasks eagerly? According to https://blog.gradle.org/preview-avoiding-task-configuration-time `withType().configureEach` configures tasks lazily which has positive impact on the build performance.

Goooler: You are right, reverted.

jeremymailen:
Thanks for keeping an eye on things. Generally I think it's useful to pin the gradle version and other settings and worry less on small details unless there's a functional difference. Will close for now.

Issue #187, opened 2021-04-28
Test failures when running on windows


jeremymailen:
The assertions do not take into account path separator differences.
See: https://github.com/jeremymailen/kotlinter-gradle/pull/184#issuecomment-820254857
h/t @Goooler

Issue #193, opened 2021-04-29
Error when running any command provided by kotlinter-gradle


LukevanTricht:
Received the dependabot PR to upgrade kotlinter-gradle from 3.4.0 to 3.4.3, and the CI for that PR is failing with these errors:

```
> Task :app:lintKotlinDebug FAILED
lint worker execution error
java.lang.NoSuchMethodError: kotlin.jvm.internal.FunctionReferenceImpl.<init>(ILjava/lang/Object;Ljava/lang/Class;Ljava/lang/String;Ljava/lang/String;I)V
	at org.jmailen.gradle.kotlinter.tasks.lint.LintWorkerAction$execute$2$lintFunc$1.<init>(LintWorkerAction.kt)
	at org.jmailen.gradle.kotlinter.tasks.lint.LintWorkerAction.execute(LintWorkerAction.kt:42)
	at org.gradle.workers.internal.DefaultWorkerServer.execute(DefaultWorkerServer.java:63)
	at org.gradle.workers.internal.NoIsolationWorkerFactory$1$1.create(NoIsolationWorkerFactory.java:67)
	at org.gradle.workers.internal.NoIsolationWorkerFactory$1$1.create(NoIsolationWorkerFactory.java:63)
	at org.gradle.internal.classloader.ClassLoaderUtils.executeInClassloader(ClassLoaderUtils.java:97)
	at org.gradle.workers.internal.NoIsolationWorkerFactory$1.lambda$execute$0(NoIsolationWorkerFactory.java:63)
	at org.gradle.workers.internal.AbstractWorker$1.call(AbstractWorker.java:44)
	at org.gradle.workers.internal.AbstractWorker$1.call(AbstractWorker.java:41)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor$CallableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:409)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor$CallableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:399)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor$1.execute(DefaultBuildOperationExecutor.java:157)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor.execute(DefaultBuildOperationExecutor.java:242)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor.execute(DefaultBuildOperationExecutor.java:150)
[...]

PR #194, opened 2021-05-02
Test kotlinter-plugin on Github Actions, fix failing tests on Windows

mateuszkwiecinski:
Fixes https://github.com/jeremymailen/kotlinter-gradle/issues/187

### Summary
What this PR does is:
1. Adds Github Actions workflow that does exactly the same as current Travis setup plus extends it by running the same flow on windows machine (to prevent https://github.com/jeremymailen/kotlinter-gradle/issues/187 happening in the future)
2. Adds new workflow that builds the plugin and applies to matrix configuration of _Gradle Version x Java Version x OS_
3. ~Adds new workflow that will add comment with the dependencies diff if plugin configuration changes - it relies on Jake Wharton's [dependency-tree-diff](https://github.com/JakeWharton/dependency-tree-diff) tool (Note: the tool is wrapped in [my personal action](https://github.com/usefulness/dependency-tree-diff-action))~
4. Fixes test failures on Windows caused by a wrong path separator
5. ~As a showcase of 3) it enforces all configurations to rely on the same kotlin depenencies version (which should address the issue @Goooler faces in #192)~

### Context
Seeing all the recent problems with backwards compatibility/platform compatibility I thought I can come up with some ideas that should make future plugin development easier by automatically ensuring all supported configurations work as expect.  

By the way, it turned out currently minimum supported Gradle version is 6.8. And from what I understand that's exactly because of the same reason as https://github.com/jeremymailen/kotlinter-gradle/issues/193. I didn't want to change that bahvior, but as far as I understand, there is no way to work around the issue when the project uses kotlin-dsl 🤷 The only global solution is to downgrade kotlin `languageVersion` to 1.3 to allow using the plugin with older Gradle Versions. 

I decided to introduce Github Actions as they seem to become a standard in open source world and with their `matrix` option they perfectly fit into our case.

I don't know if this repository has the "run actions from forks" option enabled so I don't know if it'll trigger the build, but here's how it should look like: https://github.com/mateuszkwiecinski/kotlinter-gradle/pull/2 

image

I'll try to comment out each unobvious piece of code I added here 👀

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -8,7 +8,12 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-latest, windows-latest ]
+    name: '[${{ matrix.os }}] Build plugin'
     steps:
       - uses: actions/checkout@v2
 
@@ -26,16 +31,24 @@ jobs:
       - name: Gradle Wrapper Validation
         uses: gradle/wrapper-validation-action@v1
 
-      - run: ./gradlew assemble
+      - name: ./gradlew assemble
+        uses: eskatos/gradle-command-action@v1
+        with:
+          arguments: assemble
 
-      - run: ./gradlew check
+      - name: ./gradlew check
+        uses: eskatos/gradle-command-action@v1
+        with:
+          arguments: check
 
       - name: Upload test results
         if: ${{ always() }}
         uses: actions/upload-artifact@v2
         with:
-          name: test-results
-          path: "${{ github.workspace }}/**/build/reports/tests"
+          name: test-results-${{ matrix.os }}
+          path: |
+          "${{ github.workspace }}/**/build/reports/tests"
+          "${{ github.workspace }}\**\build\reports\tests"
 
   integration-tests:
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -1,0 +1,71 @@
+name: Build Project
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: set up Java
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'adopt'
+          java-version: 16
+
+      - name: Cache .gradle
+        uses: burrunan/gradle-cache-action@v1
+        with:
+          gradle-distribution-sha-256-sum-warning: false
+
+      - name: Gradle Wrapper Validation
+        uses: gradle/wrapper-validation-action@v1
+
+      - run: ./gradlew assemble
+
+      - run: ./gradlew check
+
+      - name: Upload test results
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v2
+        with:
+          name: test-results
+          path: "${{ github.workspace }}/**/build/reports/tests"
+
+  integration-tests:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-latest, windows-latest ]
+        gradle: [ 6.8.3, current, rc ]
+        java: [ 8, 11, 15 ]
+    name: '[${{ matrix.os }}] Gradle: ${{ matrix.gradle }}, Java: ${{ matrix.java }}'
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: set up JDK
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'adopt'
+          java-version: ${{ matrix.java }}
+
+      - name: Pre-compile plugin using gradle wrapper
+        uses: burrunan/gradle-cache-action@v1
+        with:
+          job-id: integration-test-${{ matrix.java }}-${{ matrix.gradle }}
+          gradle-distribution-sha-256-sum-warning: false
+          arguments: publishToMavenLocal -Pgroup=org.jmailen.gradle.local
+
+      - uses: eskatos/gradle-command-action@v1
+        with:
+          build-root-directory: test-project
+          gradle-version: ${{ matrix.gradle }}
+          arguments: formatKotlin lintKotlin -PuseMavenLocal
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -57,15 +57,8 @@ jobs:
           distribution: 'adopt'
           java-version: ${{ matrix.java }}
 
-      - name: Pre-compile plugin using gradle wrapper
-        uses: burrunan/gradle-cache-action@v1
-        with:
-          job-id: integration-test-${{ matrix.java }}-${{ matrix.gradle }}
-          gradle-distribution-sha-256-sum-warning: false
-          arguments: publishToMavenLocal -Pgroup=org.jmailen.gradle.local
-
       - uses: eskatos/gradle-command-action@v1
         with:
           build-root-directory: test-project
           gradle-version: ${{ matrix.gradle }}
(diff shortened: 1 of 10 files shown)


.github/workflows/default.yml
```diff
@@ -0,0 +1,75 @@
...
+        os: [ ubuntu-latest, windows-latest ]
+    name: '[${{ matrix.os }}] Build plugin'
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: set up Java
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'adopt'
+          java-version: 16
```
mateuszkwiecinski, line 24: After you set fixed java toolchains, I thought why not to build the plugin on the most recent Java version?

.github/workflows/default.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: Cache .gradle
+        uses: burrunan/gradle-cache-action@v1
+        with:
+          gradle-distribution-sha-256-sum-warning: false
```
mateuszkwiecinski, line 29: That's the best action to leverage Gradle caching I know 👌 
It required to enable caching in overall (in `gradle.properties`) and that should result in using gradle cache locally as well. I think that's a good change, but if you prefer, I can enable them on the CI only

.github/workflows/default.yml
```diff
@@ -0,0 +1,75 @@
...
+        uses: eskatos/gradle-command-action@v1
+        with:
+          arguments: check
+
+      - name: Upload test results
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v2
+        with:
+          name: test-results-${{ matrix.os }}
+          path: "${{ github.workspace }}/**/build/reports/tests"
```
mateuszkwiecinski, line 49: This will make test reports easily available under `artifacts` tab:
image
They weren't available on travis, which made it a bit problematic to investigatie test failures there

.github/workflows/default.yml
```diff
@@ -0,0 +1,75 @@
...
+
+  integration-tests:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-latest, windows-latest ]
+        gradle: [ 6.8.3, current, rc ]
+        java: [ 8, 11, 15 ]
```
mateuszkwiecinski, line 58: 1. Minimum supported Gradle version is `6.8`, but I decided to set `6.8.3` as it contains number of performance fixes and I assumed if someone uses `6.8` they can easily migrate to `6.8.3`. 
2. I test on Java 15 here to keep the configuration simple. Java 16 is not supported by `current` gradle 7.0 nor by the `rc` 6.9-rc1 version. I could exclude them but I though Java 16 is not that different from java 15 so 🤷. 
I assumed that if Java 16 support becomes problematic then someone will improve the config

test-project/build.gradle.kts
```diff
@@ -0,0 +1,4 @@
+plugins {
+    kotlin("jvm") version "1.5.0"
```
mateuszkwiecinski, line 2: I'm aware there is [Compatibility](https://github.com/jeremymailen/kotlinter-gradle#compatibility) section on plugin's Readme file, but as I wasn't sure what would I want to achieve here I decided to use the most recent stable version

test-project/settings.gradle.kts
```diff
@@ -0,0 +1,8 @@
+pluginManagement {
+    includeBuild("..")
+}
+dependencyResolutionManagement {
+    repositories {
+        mavenCentral()
+    }
+}
```
mateuszkwiecinski, line 8: That's the new feature in Gradle 7.0 👌 https://docs.gradle.org/7.0/release-notes.html#plugin-development-improvements
I left a few commits that contain pre-7.0 approach (which publish the plugin to `mavenLocal` and than applies it to a test projec). I decided to _leave them_, just in case, if for some reason we can't use the new feature.

test-project/build.gradle.kts
mateuszkwiecinski: Also, I named the test app `test-project` and that's a completely separate project that includes the plugin in its buildscript. I don't feel that's the right name, but nothing better came to my mind

.github/workflows/default.yml
```diff
@@ -0,0 +1,75 @@
...
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'adopt'
+          java-version: ${{ matrix.java }}
+
+      - uses: eskatos/gradle-command-action@v1
+        with:
+          build-root-directory: test-project
+          gradle-version: ${{ matrix.gradle }}
+          arguments: formatKotlin lintKotlin
```
mateuszkwiecinski, line 75: Tasks called on the test-project are `formatKotlin` and `lintKotlin`. I didn't know how to verify the plugin has been applied correctly, how to verify if it actually works, so I decided to introduce codestyle offences to the test-project and then to run both tasks and see if the command completes. 
Sample scenarios which would result in failing workflow:
- tasks aren't registered propertly
- `formatKotlin` doesn't fix codestyle offences
- any of tasks fail due to dependency conflict (issues like: https://github.com/jeremymailen/kotlinter-gradle/issues/193)
and so on

Sample scenarios that wouldn't be caught:
- `lintKotlin` task becomes no-op
- both tasks become no-op
- custom task, based on i.e. `LintTask` doesn't work

Although, they should fail on regular, TestKit tests

build.gradle.kts
```diff
@@ -52,6 +52,14 @@ dependencies {
     testImplementation("org.jetbrains:annotations:${Versions.jetbrainsAnnotations}")
 }
 
+configurations.configureEach {
+    resolutionStrategy.eachDependency {
+        if (requested.group == "org.jetbrains.kotlin" && requested.name.startsWith("kotlin")) {
+            useVersion("1.4.32")
+        }
+    }
+}
```
mateuszkwiecinski, line 61: Depepndency diff caused by this change should be visible in the comment:
image

Related blogpost: https://developer.squareup.com/blog/surfacing-hidden-change-to-pull-requests/

mateuszkwiecinski (commented):
Ok, so as everyone can see, GA run wasn't triggered :/ 
The changes can be verified in my own fork: https://github.com/mateuszkwiecinski/kotlinter-gradle/pull/2

If you, @jeremymailen decide it is OK to start using Github Actions, you probably have to visit [Settings](https://github.com/jeremymailen/kotlinter-gradle/settings/actions) tab and enable Github Actions? Or maybe, for security reason the first setup has to be done by someone with write permissions? idk 🤷

mateuszkwiecinski:
>It's been busy lately, I might need a little time to digest.

Sure no rush! 😃

>I can probably also publish off master from a tag push using this same infrastructure

Yes, however I'm not familiar with how the version update in Readme and other places works. So that part would also have to be automated somehow. If you want me to help with that later, please do file an issue and describe the requirements there

>I'm assuming our usage level will stay in the free tier? 

Yes, Github Actions are free for open-source projects source: https://github.com/pricing

> I don't know how many contributors to this plugin actually run on windows 😂? Maybe no need for windows in the test matrix unless we anticipate some future user side bug?

I'd suggest to test on Windows as well from 2 reasons: 
- it's a low hanging fruit as it requires only one additional line in the configuration (one line to run plugin tests and one to verify user integration)
- I've already broke tests on Windows again when I tried to upgrade JUnit4 to JUnit5 😅 [failed run](https://github.com/mateuszkwiecinski/kotlinter-gradle/actions/runs/806136535)

> Transitive dependency diffs are cool! However not sure it's a high priority here?

I noticed the failure in #192 which I claim is caused by a conflict in kotlin dependencies version. That's why I came up with a solution and a suggestion how to improve future work to see _hidden_ changes as well. 
We can drop the diff coments, sure, but I'd still suggest to try them out as their configuration is separate, quite straightforward and easily removable if for some reason they don't work well here. 
I'll create a separate Pull Request with just the configuration fix and the dependency diff workflow 👍

jeremymailen (approved):
This is brilliant stuff. Very sorry I took months to review it. It's been a really busy Spring.

We might find ourselves reducing or adjusting the text matrix over time, but I think this is a good start.

PR #202, opened 2021-05-10
Fix compatibility with Gradle configuration cache

### LINKED ISSUE #200: Support configuration caching
It would be great if you could support gradle configuration caching. [The documentation](https://docs.gradle.org/current/userguide/configuration_cache.html) is pretty helpful. I'm guessing it should be pretty straightforward - for `LintTask` it's probably just a matter of injecting `ProjectLayout` and using that instead of `project.projectDir`.

Here's an example of the failure I get when running `./gradlew --configuration-cache test_app:lintKotlinMain`

```
1 problem was found reusing the configuration cache.
- Task `:test_app:lintKotlinMain` of type `org.jmailen.gradle.kotlinter.tasks.LintTask`: invocation of 'Task.project' at execution time is unsupported.
  See https://docs.gradle.org/7.0/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution
```

mateuszkwiecinski:
Fixes #200 (last commit) + fixes annoying minor build warnings

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -15,7 +15,7 @@ repositories {
 }
 
 val pluginId = "org.jmailen.kotlinter"
-val githubUrl ="https://github.com/jeremymailen/kotlinter-gradle"
+val githubUrl = "https://github.com/jeremymailen/kotlinter-gradle"
 val webUrl = "https://github.com/jeremymailen/kotlinter-gradle"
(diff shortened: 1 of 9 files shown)


jeremymailen (approved):
Thank you @mateuszkwiecinski

PR #203, opened 2021-05-10
Bump Gradle to 7.0.2

mateuszkwiecinski:
Gradle has released a new version `7.0.1` which contains number of fixes: https://github.com/gradle/gradle/milestone/173?closed=1
Gradle also released a new version `7.0.2` which contains following fixes: 
https://github.com/gradle/gradle/milestone/177?closed=1

As I've been already doing stuff around the plugin I thought I can bump Gradle as well, ~but to save everyone's time in the future I also suggest to use a Github Action that will automate that process.~

~Feel free to revert the last commit if you prefer bumping things manually.~

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -84,7 +84,7 @@ tasks {
     }
 
     wrapper {
-        gradleVersion = "7.0"
+        gradleVersion = "7.0.2"
     }
 }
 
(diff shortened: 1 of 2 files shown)


jeremymailen:
We probably would prefer to upgrade gradle manually.

jeremymailen (approved):
Thank you sir!

PR #211, opened 2021-07-19
GitHub Action for automatically publishing plugin on release

jeremymailen:
A more mature CI suitable to multiple collaborators.

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -13,11 +13,11 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, windows-latest ]
-    name: '[${{ matrix.os }}] Build plugin'
+    name: '[${{ matrix.os }}] build plugin'
     steps:
       - uses: actions/checkout@v2
 
-      - name: set up Java
+      - name: Set up Java
         uses: actions/setup-java@v2
         with:
           distribution: 'adopt'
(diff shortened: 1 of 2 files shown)


jeremymailen:
@mateuszkwiecinski I believe this should work, but let me know if you have any suggestions.

mateuszkwiecinski:
I can confirm it should work in a way it will build the jars and upload them to gradle plugin portal 👍, but what about the plugin version? Where the published artifact's version will come from? Do you have to bump versions in a separate commit beforehand? 

I personally use https://github.com/allegro/axion-release-plugin to store project version in git tags, so when creating a release it is created with what I just passed.  
The disadvantage is: a version tag is created before the plugin is available on maven repository, but I recall Jake Wharton created https://github.com/JakeWharton/dependency-watch so I assume he turns things around and it becomes possible to trigger a build (i.e. from github actions UI using `workflow_dispatch` event), wait until it is becomes available for everyone and finally to create a git tag 👀

jeremymailen:
Thank you for the tips. I think for now it's simple enough to continue to bump the release version in a PR. Parts of the readme also need to be edited to go with it -- which I'm sure can be templated, but I always value simplicity. Typically I post the new release within a minute or two of landing the release PR, so the gap probably isn't going to trip anyone up.

PR #215, opened 2021-07-31
Upgrade ktlint -> 0.42.0

Mahoney:
Fixes a nasty performance bug: https://github.com/pinterest/ktlint/issues/1135

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -32,6 +32,20 @@ object Versions {
 }
 
 dependencies {
+
+    listOf(
+        "kotlin-compiler-embeddable",
+        "kotlin-stdlib",
+        "kotlin-stdlib-jdk8",
+    ).forEach { module ->
+        api("org.jetbrains.kotlin:$module") {
+            version {
+                strictly("[1.4.0, 1.5.0[")
+                prefer("1.4.32")
+            }
+        }
+    }
+
     compileOnly("org.jetbrains.kotlin:kotlin-gradle-plugin")
     compileOnly("com.android.tools.build:gradle:${Versions.androidTools}")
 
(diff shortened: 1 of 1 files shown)


Mahoney:
Wow. Didn't expect that to break the world!

Mahoney:
Ugh... ktlint has bumped the kotlin version to 1.5.20. So can't upgrade without gradle upgrading their embedded kotlin version.

gradle depending on a specific version of kotlin, and being so slow to update, is a grade a pain.

Mind you, ktlint going up a major version of kotlin in a release that fixes a major performance regression is also less than kind.

Mahoney:
Build passes if I fix the kotlin version.

Mahoney:
Nope, doesn't then work as a plugin. Fail.

jeremymailen:
Hi @Mahoney thanks for the assist. I've been tracking the progress of 0.42.0 and will upgrade soon. Right now it looks like it's got some indent regressions, so I'm waiting in anticipation of a 0.42.1 soon which might address them.

Unless something new has gone sideways with gradle (such as it did before 6.8) it should be possible to take the 1.5.x kotlin upgrade even if have to declare the newest version of kotlinter will require kotlin 1.5.x. This happens periodically because of the dependency on the internal kotlin compiler.

BTW, I do not believe pinterest/ktlint#1135 effects kotlinter. We integrate directly with the `ktlint-core` portion of the library and do not use the file globbing code in the cli component where the performance issue was. Are you seeing performance problems in kotlinter?

Mahoney:
OK, not scientific but having installed the pre push commit hook I had had a couple of times when it felt very slow, saw the performance regression on ktlint and assumed that was it.

It might be that gradle keeps having to spin up a new daemon because of a memory leak somewhere in my build.

PR #229, opened 2021-12-17
Fix `src/main/kotlin` not being recognised by default

### LINKED ISSUE #228: Doesn't work in `src/main/kotlin`
After AGP 7.0 we can use `src/main/kotlin` as kotlin sourceSet by default, but the kotlinter doesn't work in this location.

See JLLeitschuh/ktlint-gradle#524.

mateuszkwiecinski:
Fixes #228

Seeing the linked issue, I was curious if that's really the case 👀 
I wanted to reproduce the issue in tests, which led me to bumping things to make the whole project be compiled against AGP 7.X and to basically make the project up-to-date again. 
Not sure if you like the changes I'm proposing here. I tried to keep them in separate commits, just in case you want to take only some of them. 

Important note, the changes requires consumers to use AGP 7.0 and up, due to usage of the `.kotlin` extension. If the goal is to maintain old Gradle compatibility then I guess it's required to access kotlin sourcesets in a smarter way)

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -50,7 +50,7 @@ jobs:
       matrix:
         os: [ ubuntu-latest, windows-latest ]
         gradle: [ 7.0.1, current, rc ]
-        java: [ 8, 11, 16 ]
+        java: [ 8, 11, 17 ]
     name: '[${{ matrix.os }}] Gradle: ${{ matrix.gradle }}, Java: ${{ matrix.java }}'
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -49,7 +49,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, windows-latest ]
-        gradle: [ 7.0.1, current, rc ]
+        gradle: [ 7.0.1, current, release-candidate ]
         java: [ 8, 11, 17 ]
         exclude:
           - os: ubuntu-latest
@@ -76,3 +76,37 @@ jobs:
           build-root-directory: test-project
           gradle-version: ${{ matrix.gradle }}
           arguments: formatKotlin lintKotlin
+
+  integration-tests-android:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        gradle: [ current ]
+        java: [ 17 ]
+        agp: [ 7.0.4 ]
+        include:
+          - gradle: 7.0.1
+            java: 11
+            agp: 4.2.2
+          - gradle: 6.9.1
+            java: 8
+            agp: 4.0.0
+
+    name: '[android] Gradle: ${{ matrix.gradle }}, Java: ${{ matrix.java }}, AGP: ${{ matrix.agp }}'
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Set up JDK
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'adopt'
+          java-version: ${{ matrix.java }}
+
+      - uses: gradle/gradle-build-action@v2
+        with:
+          build-root-directory: test-project
+          gradle-version: ${{ matrix.gradle }}
+          arguments: formatKotlin lintKotlin -PagpVersion=${{ matrix.agp }}
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -107,6 +107,6 @@ jobs:
 
       - uses: gradle/gradle-build-action@v2
         with:
-          build-root-directory: test-project
+          build-root-directory: test-project-android
           gradle-version: ${{ matrix.gradle }}
           arguments: formatKotlin lintKotlin -PagpVersion=${{ matrix.agp }}
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -51,6 +51,14 @@ jobs:
         os: [ ubuntu-latest, windows-latest ]
         gradle: [ 7.0.1, current, rc ]
         java: [ 8, 11, 17 ]
+        exclude:
+          - os: ubuntu-latest
+            gradle: 7.0.1
+            java: 17
+          - os: windows-latest
+            gradle: 7.0.1
+            java: 17
+
     name: '[${{ matrix.os }}] Gradle: ${{ matrix.gradle }}, Java: ${{ matrix.java }}'
     steps:
       - uses: actions/checkout@v2
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,3 +1,4 @@
+import org.jetbrains.kotlin.gradle.plugin.getKotlinPluginVersion
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
@@ -5,7 +6,7 @@ plugins {
     id("com.gradle.plugin-publish") version "0.18.0"
     `java-gradle-plugin`
     `maven-publish`
-    id("org.jmailen.kotlinter") version "3.6.0"
+    id("org.jmailen.kotlinter") version "3.7.0"
     idea
 }
 
@@ -24,10 +25,23 @@ group = "org.jmailen.gradle"
 description = projectDescription
 
 object Versions {
-    const val androidTools = "4.2.2"
+    const val androidTools = "7.0.4"
     const val junit = "4.13.2"
-    const val ktlint = "0.43.0"
+    const val ktlint = "0.43.2"
     const val mockitoKotlin = "4.0.0"
+}
+
+configurations {
+    register("testRuntimeDependencies") {
+        extendsFrom(compileOnly.get())
+    }
+    configureEach {
+        resolutionStrategy.eachDependency {
+            if (requested.group == "org.jetbrains.kotlin" && requested.name.startsWith("kotlin")) {
+                useVersion(getKotlinPluginVersion())
+            }
+        }
+    }
 }
 
 dependencies {
@@ -53,7 +67,7 @@ dependencies {
 
 java {
     toolchain {
-        languageVersion.set(JavaLanguageVersion.of(8))
+        languageVersion.set(JavaLanguageVersion.of(JavaVersion.VERSION_17.majorVersion))
     }
 }
 
@@ -69,17 +83,21 @@ tasks {
         dependsOn(generateVersionProperties)
     }
 
+    val targetJavaVersion = JavaVersion.VERSION_1_8
+    withType<JavaCompile>().configureEach {
+        options.release.set(targetJavaVersion.majorVersion.toInt())
+    }
     withType<KotlinCompile>().configureEach {
         kotlinOptions {
             apiVersion = "1.4"
             languageVersion = "1.4"
-            jvmTarget = JavaVersion.VERSION_1_8.toString()
+            jvmTarget = targetJavaVersion.toString()
         }
     }
 
     // Required to put the Kotlin plugin on the classpath for the functional test suite
     withType<PluginUnderTestMetadata>().configureEach {
-        pluginClasspath.from(configurations.compileOnly)
+        pluginClasspath.from(configurations.getByName("testRuntimeDependencies"))
     }
 
     wrapper {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -83,7 +83,7 @@ tasks {
     }
 
     wrapper {
-        gradleVersion = "7.3.1"
+        gradleVersion = "7.3.2"
     }
 }
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -85,7 +85,7 @@ tasks {
     }
 
     wrapper {
-        gradleVersion = "7.2"
+        gradleVersion = "7.3.1"
     }
 }
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,8 +1,8 @@
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
-    kotlin("jvm") version "1.5.31"
-    id("com.gradle.plugin-publish") version "0.15.0"
+    kotlin("jvm") version "1.6.10"
+    id("com.gradle.plugin-publish") version "0.18.0"
     `java-gradle-plugin`
     `maven-publish`
     id("org.jmailen.kotlinter") version "3.6.0"
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -25,7 +25,6 @@ description = projectDescription
 
 object Versions {
     const val androidTools = "4.2.2"
-    const val jetbrainsAnnotations = "22.0.0"
     const val junit = "4.13.2"
     const val ktlint = "0.43.0"
     const val mockitoKotlin = "4.0.0"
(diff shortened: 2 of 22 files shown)


.github/workflows/default.yml
```diff
@@ -49,8 +49,16 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, windows-latest ]
-        gradle: [ 7.0.1, current, rc ]
-        java: [ 8, 11, 16 ]
+        gradle: [ 7.3, current, rc ]
+        java: [ 8, 11, 17 ]
```
mateuszkwiecinski, line 53: I thought it makes sense to test on Java 17. I expect more people start using it as it is an LTS version

build.gradle.kts
```diff
@@ -24,13 +25,25 @@ group = "org.jmailen.gradle"
...
     const val junit = "4.13.2"
-    const val ktlint = "0.43.0"
+    const val ktlint = "0.43.2"
     const val mockitoKotlin = "4.0.0"
 }
 
+configurations {
+    register("testRuntimeDependencies") {
+        extendsFrom(compileOnly.get())
+    }
```
mateuszkwiecinski, line 37: `compileOnly` isn't resolvable anymore. I explicitly declare new configuration which can be used via `pluginUnderTestMetadata` task

build.gradle.kts
```diff
@@ -50,12 +63,11 @@ dependencies {
 
     testImplementation("junit:junit:${Versions.junit}")
     testImplementation("org.mockito.kotlin:mockito-kotlin:${Versions.mockitoKotlin}")
-    testImplementation("org.jetbrains:annotations:${Versions.jetbrainsAnnotations}")
 }
 
 java {
     toolchain {
-        languageVersion.set(JavaLanguageVersion.of(8))
+        languageVersion.set(JavaLanguageVersion.of(JavaVersion.VERSION_17.majorVersion))
```
mateuszkwiecinski, line 70: I didn't have Java 8  installed, so I thought it makes sense to set recent LTS to be built with, but still target Java 8 (the `options.release` call below)

src/main/kotlin/org/jmailen/gradle/kotlinter/pluginapplier/AndroidSourceSetApplier.kt
```diff
@@ -14,18 +14,19 @@ import org.jmailen.gradle.kotlinter.id
...
         }
     }
 
-    private fun getKotlinFiles(project: Project, sourceSet: AndroidSourceSet) = sourceSet.java.srcDirs.map { dir ->
-        project.fileTree(dir) { it.include("**/*.kt") }
-    }.reduce { merged: FileTree, tree: ConfigurableFileTree ->
-        merged + tree
+    private fun getKotlinFiles(project: Project, sourceSet: AndroidSourceSet): FileTree? {
+        val javaSources = sourceSet.java.srcDirs
+        val kotlinSources = (sourceSet.kotlin as? com.android.build.gradle.api.AndroidSourceDirectorySet)?.srcDirs.orEmpty()
```
mateuszkwiecinski, line 26: `sourceSet.kotlin` is accessible only on AGP 7.0 and up. AFAIU this wil fail if consumer's project uses lower version. This line is also a fix for the linked issue.

src/main/kotlin/org/jmailen/gradle/kotlinter/pluginapplier/KotlinJvmSourceSetApplier.kt
```diff
@@ -11,7 +11,7 @@ import org.jmailen.gradle.kotlinter.id
 internal object KotlinJvmSourceSetApplier : SourceSetApplier {
 
     override fun applyToAll(project: Project, action: SourceSetAction) {
-        getSourceSets(project).all { sourceSet ->
+        getSourceSets(project).configureEach { sourceSet ->
```
mateuszkwiecinski, line 14: As a boy-scout rule, I replaced non-lazy apis usages with lazy ones

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/AndroidProjectTest.kt
```diff
@@ -0,0 +1,101 @@
...
+                }
+                resolve("src/debug/kotlin/DebugSourceSet.kt") {
+                    writeText(kotlinClass("DebugSourceSet"))
+                }
+                resolve("src/test/kotlin/TestSourceSet.kt") {
+                    writeText(kotlinClass("TestSourceSet"))
+                }
+                resolve("src/flavorOne/kotlin/FlavorSourceSet.kt") {
+                    writeText(kotlinClass("FlavorSourceSet"))
+                }
```
mateuszkwiecinski, line 79: If needed, one can also add test case fo `src/xxx/java`, to verify `/java` directory is also recognised properly

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/CustomTaskTest.kt
```diff
@@ -22,7 +21,7 @@ class CustomTaskTest : WithGradleTest.Kotlin() {
         projectRoot = testProjectDir.root.apply {
             resolve("settings.gradle") { writeText(settingsFile) }
             resolve("build.gradle") {
-                @Language("groovy")
+                // language=groovy
```
mateuszkwiecinski, line 24: I also noticed that I unnecessairly added a dependency, which can be easily replaced with comments

build.gradle.kts
```diff
@@ -24,13 +25,25 @@ group = "org.jmailen.gradle"
...
+        extendsFrom(compileOnly.get())
+    }
+    configureEach {
+        resolutionStrategy.eachDependency {
+            if (requested.group == "org.jetbrains.kotlin" && requested.name.startsWith("kotlin")) {
+                useVersion(getKotlinPluginVersion())
+            }
+        }
+    }
+}
```
mateuszkwiecinski, line 45: Ah, forgot to mention that one. That's basically the same as in #201, with the same reasoning, but this time _something_ in tests config uses incompatible version of kotlin, so instead of digging which dependency needs to be bumped I'm propsoing to simply keep them consistent 🤷‍♂️

.github/workflows/default.yml
jeremymailen: Agree

build.gradle.kts
jeremymailen: Sounds good

.github/workflows/default.yml
```diff
@@ -50,7 +50,15 @@ jobs:
       matrix:
         os: [ ubuntu-latest, windows-latest ]
         gradle: [ 7.0.1, current, rc ]
-        java: [ 8, 11, 16 ]
+        java: [ 8, 11, 17 ]
```
jeremymailen, line 53: Do we also need to update the release.yml workflow to use Java 17?

build.gradle.kts
```diff
@@ -24,13 +25,25 @@ group = "org.jmailen.gradle"
 description = projectDescription
 
 object Versions {
-    const val androidTools = "4.2.2"
-    const val jetbrainsAnnotations = "22.0.0"
+    const val androidTools = "7.0.4"
     const val junit = "4.13.2"
-    const val ktlint = "0.43.0"
+    const val ktlint = "0.43.2"
```
jeremymailen, line 30: Yeah, it's probably good to update. I hadn't yet because afaik nothing in 0.43.1 and 0.43.2 impacted kotlinter.
Still waiting on a fix to wildcard imports :/.

build.gradle.kts
```diff
@@ -24,13 +25,25 @@ group = "org.jmailen.gradle"
 description = projectDescription
 
 object Versions {
-    const val androidTools = "4.2.2"
-    const val jetbrainsAnnotations = "22.0.0"
+    const val androidTools = "7.0.4"
```
jeremymailen, line 28: Yeah, there's probably no way around losing compatibility with prior versions of AGP since we use the plugin itself to resolve the source sets and can't easily include two in the compile.

Can you go ahead and add an Android Plugin column to the compatibility matrix?
https://github.com/jeremymailen/kotlinter-gradle#compatibility

src/main/kotlin/org/jmailen/gradle/kotlinter/pluginapplier/KotlinJvmSourceSetApplier.kt
jeremymailen: Yes, in fact we had needed to be lazy in a few places in the code because if someone customizes the source sets in their build file we need to resolve our targets after that is configured, otherwise in early versions of kotlinter we didn't pick up the customization. I assume this is harmonious with that?

src/main/kotlin/org/jmailen/gradle/kotlinter/pluginapplier/AndroidSourceSetApplier.kt
jeremymailen: Yes, I guess we'll need a compatibility boundary as per my earlier comment.

jeremymailen (commented):
Thank you as always for your diligence in watching over the project!

This looks good to go with the Readme update and release workflow bump. Thank you!

src/main/kotlin/org/jmailen/gradle/kotlinter/pluginapplier/KotlinJvmSourceSetApplier.kt
mateuszkwiecinski: AFAIR The old issue was about customization of an existing sourceset not being picked up. This change will "fix" the plugin if someone _registers new_ sourceset after kotlinter plugin gets applied. So yeah, it's definitelly consistent with the other behavior 👍

.github/workflows/default.yml
mateuszkwiecinski: Good catch 👍 I clearly missed that!

build.gradle.kts
mateuszkwiecinski: I pushed another commit, with a "fix" which would maintain old behavior for AGP's `4.x` users. I also added `integration-tests-android` job to existing workflow which makes sure the plugin will work with older AGP versions.

image

src/main/kotlin/org/jmailen/gradle/kotlinter/pluginapplier/AndroidSourceSetApplier.kt
```diff
@@ -14,18 +14,21 @@ import org.jmailen.gradle.kotlinter.id
...
 
-    private fun getKotlinFiles(project: Project, sourceSet: AndroidSourceSet) = sourceSet.java.srcDirs.map { dir ->
-        project.fileTree(dir) { it.include("**/*.kt") }
-    }.reduce { merged: FileTree, tree: ConfigurableFileTree ->
-        merged + tree
+    private fun getKotlinFiles(project: Project, sourceSet: AndroidSourceSet): FileTree? {
+        val javaSources = sourceSet.java.srcDirs
+        val kotlinSources = runCatching { (sourceSet.kotlin as? com.android.build.gradle.api.AndroidSourceDirectorySet)?.srcDirs }
+            .getOrNull()
+            .orEmpty()
```
mateuszkwiecinski, line 28: This is a _trick_ which I read as "if there is a speparate kotlin sourceset append it to java ones". I initially wanted to do a version check on AGP, but I couldn't find a stable api to do so. 
I once reported: https://issuetracker.google.com/issues/188859855 where I tried to let AGP maintainers know it's unclear what's the preferred way to obtain AGP version. 
The alternatives I considered are:
- add their internal artifact and use Version object
- continue using deprecated property until they remove it at some point
- do a reflective call to check if `kotlin` property exists

As a conclusion I'm proposing to do catch and ignore the exception if AGP 7.0 specific classes aren't available in runtime

build.gradle.kts
jeremymailen: Very nice

src/main/kotlin/org/jmailen/gradle/kotlinter/pluginapplier/AndroidSourceSetApplier.kt
jeremymailen: That seems like a reasonable approach all considered. This or that.

.github/workflows/default.yml
```diff
@@ -68,3 +76,37 @@ jobs:
...
+
+      - name: Set up JDK
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'adopt'
+          java-version: ${{ matrix.java }}
+
+      - uses: gradle/gradle-build-action@v2
+        with:
+          build-root-directory: test-project
```
jeremymailen, line 110:
```suggestion
          build-root-directory: test-project-android
```

.github/workflows/default.yml
```diff
@@ -68,3 +76,37 @@ jobs:
...
+        gradle: [ current ]
+        java: [ 17 ]
+        agp: [ 7.0.4 ]
+        include:
+          - gradle: 7.0.1
+            java: 11
+            agp: 4.2.2
+          - gradle: 6.9.1
+            java: 8
+            agp: 4.0.0
```
jeremymailen, line 94: This version isn't found. Perhaps it requires a different repo or a different patch version?
I'm fine dropping that from the matrix if it end up being a pain.

src/main/kotlin/org/jmailen/gradle/kotlinter/pluginapplier/AndroidSourceSetApplier.kt
```diff
@@ -14,18 +14,21 @@ import org.jmailen.gradle.kotlinter.id
...
         }
     }
 
-    private fun getKotlinFiles(project: Project, sourceSet: AndroidSourceSet) = sourceSet.java.srcDirs.map { dir ->
-        project.fileTree(dir) { it.include("**/*.kt") }
-    }.reduce { merged: FileTree, tree: ConfigurableFileTree ->
-        merged + tree
+    private fun getKotlinFiles(project: Project, sourceSet: AndroidSourceSet): FileTree? {
+        val javaSources = sourceSet.java.srcDirs
+        val kotlinSources = runCatching { (sourceSet.kotlin as? com.android.build.gradle.api.AndroidSourceDirectorySet)?.srcDirs }
```
jeremymailen, line 26: Let's put a comment above this line noting why. Something like "detect kotlin source paths supported by AGP 7 and later"

jeremymailen (commented):
NIcely done. A couple of fixes.
Once we land this I'll probably push a minor release.

.github/workflows/default.yml
mateuszkwiecinski: Yeah, I wasn't sure what's the lower boundary so I picked a version I consider absolute minimum and I would be surprised if any continously maintained project used lower version. I know some project refrain from migrating to 7.0 because it still has some hickups and I _heard_ the 7.2 should be considered the "proper" version of AGP 7 (7.2 is still in alpha and is planned for mid February) 

My initial setup didn't work because older AGP versions don't publish gradle metadata needed for `plugins` block ([Issuetracker](https://issuetracker.google.com/issues/64551265)). I dropped `plugins`  block in favor of custom `resolutionStrategy`[here](https://github.com/jeremymailen/kotlinter-gradle/pull/229/files#diff-124a21ab4c45e752a2d3151e279332a337c772e9198fdd248dd43a1127c3e6c9R8)

>I'm fine dropping that from the matrix if it end up being a pain.

It wasn't that big pain so I decided to include necessary changes. My motivation was to explicitly support 4.0 to see the workflow failing when some changes in Kotlinter plugin will introduce breaking changes, so it can be mentioned in the release notes

jeremymailen (approved):

JLLeitschuh:
PRs are welcome to resolve this issue, please do include test coverage.

PR #238, opened 2022-02-21
Support KtLint 0.44.0

lwasyl:


lwasyl:
I also bumped version since I imagine you may want to release support for the new ktlint version

jeremymailen (approved):
Thank you!

PR #241, opened 2022-03-19
Bump Gradle 7.4.1, fix `generateVersionProperties` configuration-cache compatibility

### LINKED ISSUE #227: Kotlinter 3.7.0 removes needed star imports
Kotlinter removes the star import here, which makes our builds fail:

```
import com.example.processor.*
import com.example.processor.LocationType.CITY
import com.example.processor.LocationType.COUNTRY
```
Here's `LocationType`:
```
enum class LocationType {
    CITY,
    COUNTRY;
}
```
I refer to `CITY`, `COUNTRY` and `LocationType` in the calling code.

mateuszkwiecinski:
I wanted to bump ktlint version to fix/ #227, but I stumbled upon https://github.com/pinterest/ktlint/issues/1421 which turned out to be a blocker :/ 

While investigating the issue, I noticed a warning/error saying:
>1 problem was found storing the configuration cache.
>- Task `:generateVersionProperties` of type `org.gradle.api.DefaultTask`: cannot serialize Gradle script object references, as these are not supported with the configuration cache.
  See https://docs.gradle.org/7.3.2/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
  
 
In this PR I'm proposing to:
 1. Bump Gradle to _current_ version
 2. Bump internal kotlinter dependency and run `formatKotlin` to have all offences autocorrected
 3. Declare inputs for `generateVersionProperties` task, so the build is compatible with configuration-cache
 4. Alter `generateVersionProperties` behavior and generate the properties file directly under `build` directory


build.gradle.kts
```diff
@@ -73,9 +73,13 @@ java {
 
 tasks {
     val generateVersionProperties = register("generateVersionProperties") {
+        val projectVersion = version
+        val propertiesFile = File(sourceSets.main.get().output.resourcesDir, "version.properties")
+        inputs.property("projectVersion", projectVersion)
+        outputs.file(propertiesFile)
```
mateuszkwiecinski, line 79: I couldn’t find a reason why modify the project sources was needed, so I thought I can propose to modify the `build` directory directly. 
I verified the file generates under the same location:

image

and it can be found in published jar:
image

If my assumptions are invalid, previous behaviour would be:
```suggestion
        val projectVersion = version
        val propertiesFile = File(sourceSets.main.get().resources.sourceDirectories.asPath, "version.properties")
        inputs.property("projectVersion", projectVersion)
        outputs.file(propertiesFile)
```

jeremymailen (approved):
Nice! Good improvement

PR #245, opened 2022-04-02
Drop 'indentSize'

mateuszkwiecinski:
From the discussion in https://github.com/pinterest/ktlint/issues/1434#issuecomment-1086621945 I understood overriding an `.editorConfig` property via `userData` parameter won't be supported.
According to https://github.com/pinterest/ktlint/pull/1442 passing `indent_size` will result in a runtime exception. Regardless if this lands in current form, in context of the discussion from [previous PR](https://github.com/jeremymailen/kotlinter-gradle/pull/243#issuecomment-1074834159) - I'm proposing to remove the `indentSize` property from Kotlinter's extension.
Regardless of how the future of ktlint will look like, I think such change _makes sense_:
a) `.editorconfig` should be the preferred way of setting the indent_size parameter, mainly because it integrates with IntelliJ based IDEs.
b) it's less code to maintain 😛 

The recommended migration path which probably should be mentioned in release notes is to replace `kotlinter.indentSize` with the following entry in `{project_root}/.editorconfig`:
```editorconfig
[*.{kt, kts}]
indent_size = 4
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,6 @@ Options are configured in the `kotlinter` extension. Defaults shown (you may omi
 ```kotlin
 kotlinter {
     ignoreFailures = false
-    indentSize = 4
     reporters = arrayOf("checkstyle", "plain")
     experimentalRules = false
     disabledRules = emptyArray()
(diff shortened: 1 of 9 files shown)


jeremymailen (approved):
I agree completely. The indentSize in the gradle extension is vestigial at this point.
The documentation will be a little ahead of the release, but it's ok I think since we'll get the upgrade done this week.

Issue #186, opened 2021-04-26
[Question] How to add custom rules via plugins block?


a-ivanov:
Hello everyone!

If I install the kotlinter via Gradle `plugins` block

```
plugins {
    id("org.jmailen.kotlinter") version "<version>"
}
```

is it possible to add my custom rules (e.g. from another jar)?

I know that `buildscript` block option is described in Readme, but afaik this is not compatible with `plugins` block.

Thanks.

jeremymailen:
Correct, I believe gradle requires you to use the `buildscript` block when you need to modify the classpath of the build and that's where you will need to include your custom rules in the classpath. Hope this helps.

Issue #34, opened 2018-01-12
Is that possible to include custom rule from current project?


MyDogTom:
Readme states two ways how to add custom rules
```gradle
buildscript {
    dependencies {
        classpath files('libs/my-custom-ktlint-rules.jar')
        classpath 'org.other.ktlint:custom-rules:1.0'
    }
}
```

I'd like to have custom rules as separate module in my project. Is that possible to include them directly? Ideally, I'd like that these rules are applied against themselves as well.

jeremymailen:
kotlinter will resolve any lint rules in the classpath of the gradle buildscript. However usually you can only use dependencies in `buildscript` which can resolved during gradle's evaluation phase which would exclude anything that gradle build produces. Someone had a similar question here: https://discuss.gradle.org/t/code-generation-problems-buildsrc-and-buildscript-dependency-on-another-module/5681
Chicken and egg problem. You'd probably need a separate gradle build which publishes to mavenLocal or something. It could validate itself against an older published version of itself.

jeremymailen:
Since I don't think there's anything straightforward we could do in kotlinter to work around this constraint in gradle, I'll close this for now. Thanks.

Kritarie:
+1 We currently have a clunky process where we build our custom rules jar, copy it over to a /libs dir (checked into source control), then use buildscript { classpath(...) } to include them. Would rather do what we do for lint rules:
```
dependencies {
  lintChecks(project(":custom-checks"))
}
```

jeremymailen:
Let me reopen. There's a PR out proposing this. I think it will need some careful testing, but I might have become paranoid about gradle workers :).

eric239:
@jeremymailen pinging this thread... do you think this can be fixed any time soon?

jeremymailen:
It might take a little bit, since the upgrade to ktlint 0.47.x is up next.
This would be a bigger change and require careful consideration.

jeremymailen:
At long last this is supported in [5.0.0](https://github.com/jeremymailen/kotlinter-gradle/releases/tag/5.0.0) by adding your rules project to the `ktlint` dependency.

PR #250, opened 2022-04-22
introduce ktlintRuleSet configuration (BC break!)

hrach:
The PR brings an easy way to add custom ktlint rules.

Why it is a difficult now:
- There is no way to include other project's module as ktlint ruleset source. https://github.com/jeremymailen/kotlinter-gradle/issues/34
- Using a buildscript is in contradiction with plugins block (a preferred way how to configure project nowadays) and it seems not to be possible to combine them (I'm not fully sure but it didn't worked for me and this seems to be [confirmed here](https://github.com/jeremymailen/kotlinter-gradle/issues/186).)

So overall I've applied the same approach here as it is used in a competitive gradle ktlint plugin we were using previously before switching to this one.

### DIFF
--- a/src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
+++ b/src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
@@ -3,6 +3,8 @@ package org.jmailen.gradle.kotlinter
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
+import org.gradle.api.artifacts.Configuration
+import org.gradle.api.attributes.Bundling
 import org.gradle.api.tasks.TaskProvider
 import org.jmailen.gradle.kotlinter.pluginapplier.AndroidSourceSetApplier
 import org.jmailen.gradle.kotlinter.pluginapplier.KotlinJvmSourceSetApplier
@@ -15,6 +17,15 @@ import org.jmailen.gradle.kotlinter.tasks.LintTask
 import java.io.File
 
 class KotlinterPlugin : Plugin<Project> {
+
+    private object Versions {
+        const val ktlint = "0.45.2"
+    }
+
+    companion object {
+        private const val KTLINT_CONFIGURATION_NAME = "ktlint"
+        private const val RULE_SET_CONFIGURATION_NAME = "ktlintRuleSet"
+    }
 
     private val extendablePlugins = mapOf(
         "org.jetbrains.kotlin.jvm" to KotlinJvmSourceSetApplier,
@@ -36,6 +47,9 @@ class KotlinterPlugin : Plugin<Project> {
                 val lintKotlin = registerParentLintTask()
                 val formatKotlin = registerParentFormatTask()
 
+                val ktlintConfiguration = createKtlintConfiguration()
+                val ruleSetConfiguration = createRuleSetConfiguration(ktlintConfiguration)
+
                 sourceResolver.applyToAll(this) { id, resolvedSources ->
                     val lintTaskPerSourceSet = tasks.register("lintKotlin${id.capitalize()}", LintTask::class.java) { lintTask ->
                         lintTask.source(resolvedSources)
@@ -49,6 +63,8 @@ class KotlinterPlugin : Plugin<Project> {
                         )
                         lintTask.experimentalRules.set(provider { kotlinterExtension.experimentalRules })
                         lintTask.disabledRules.set(provider { kotlinterExtension.disabledRules.toList() })
+                        lintTask.ktlintClasspath.setFrom(ktlintConfiguration)
+                        lintTask.ruleSetsClasspath.setFrom(ruleSetConfiguration)
                     }
                     lintKotlin.configure { lintTask ->
                         lintTask.dependsOn(lintTaskPerSourceSet)
@@ -59,6 +75,8 @@ class KotlinterPlugin : Plugin<Project> {
                         formatTask.report.set(reportFile("$id-format.txt"))
                         formatTask.experimentalRules.set(provider { kotlinterExtension.experimentalRules })
                         formatTask.disabledRules.set(provider { kotlinterExtension.disabledRules.toList() })
+                        formatTask.ktlintClasspath.setFrom(ktlintConfiguration)
+                        formatTask.ruleSetsClasspath.setFrom(ruleSetConfiguration)
                     }
                     formatKotlin.configure { formatTask ->
                         formatTask.dependsOn(formatKotlinPerSourceSet)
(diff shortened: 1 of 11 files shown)


src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
```diff
@@ -16,6 +18,15 @@ import java.io.File
 
 class KotlinterPlugin : Plugin<Project> {
 
+    private object Versions {
+        const val ktlint = "0.45.2"
```
hrach, line 22: Ideally this should be somehow shared with the build script.

jeremymailen:
Thank you @hrach! I'll take a look this week -- thanks for your patience -- just got back from vacation.

src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
jeremymailen: Would it prevent customizing the ktlint version?

src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
```diff
@@ -93,6 +111,31 @@ class KotlinterPlugin : Plugin<Project> {
...
         }
+
+    private fun Project.createKtlintConfiguration(): Configuration = configurations.maybeCreate(KTLINT_CONFIGURATION_NAME).apply {
+        isCanBeResolved = true
+        isCanBeConsumed = false
+        isVisible = false
+
+        val dependencyProvider = provider {
+            this@createKtlintConfiguration.dependencies.create(
+                "com.pinterest:ktlint:${Versions.ktlint}"
```
jeremymailen, line 122: Any way to reference a BOM or something like that?

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/FormatTask.kt
```diff
@@ -26,26 +28,40 @@ open class FormatTask @Inject constructor(
...
     @TaskAction
     fun run() {
-        val result = with(workerExecutor.noIsolation()) {
-            submit(FormatWorkerAction::class.java) { p ->
-                p.name.set(name)
-                p.files.from(source)
-                p.projectDirectory.set(projectLayout.projectDirectory.asFile)
-                p.ktLintParams.set(getKtLintParams())
-                p.output.set(report)
+        val workQueue = workerExecutor.processIsolation { spec ->
```
jeremymailen, line 43: Are you sure this isolation level works? I remember trying various ones in order to try to get a classloader isolation for the version of kotlin used by ktlint (because it uses internal APIs) to wall it off from the version the user may have configured, but in the end I had a hard time getting it working.
If it does work, is there any performance impact?

jeremymailen (commented):
Thank you! Love that you included tests and were thorough in the details.

I think I get the gist -- we're making a more standardized dependencies configuration the rules --  but could you add a few sentences in the PR describing the goals and thinking behind the PR? Plus any notes on testing it.

It looks like a couple of the automated tests are failing.

hrach:
> could you add a few sentences in the PR describing the goals and thinking behind the PR?

done

> notes on testing it

Testing is done the same way other stuff is tested here but I find it a bit fragile as current test are teting that automatic fixes are able there and working. But if the issue is not detected at all, test still passes. But maybe you're asking about somethign else.(?)

> It looks like a couple of the automated tests are failing.

Yeah, I will investigate.

jeremymailen:
Thanks for your patience -- unusually busy time for me right now in my day job.

Thank you for the description -- totally makes sense -- you are adding new capability, not just syntactic sugar.

I ask about testing more because there might be additional compatibility or performance test cases to consider.

My prior experience with Gradle Workers -- when it was a less mature API -- was that different isolation levels could cause problems. At one point I tried to use them to isolate the version of kotlin that ktlint pulled in from the version which the consumer of the plugin depended on for their build. In the end I had to abandon the effort because it either caused unusual timing exceptions transfering results back to a shared log file or made the plugin really slow. Since we parcel out lots of groups of files as tasks to workers to maximize parallel execution, running each task in a separate process or classloader could really impact performance -- which is a core tenant of kotlinter: performance and simplicity.

What I'll do is pull down this version and test it on a few sizable projects to see if performance is impacted or any parallel execution bugs crop up.

hrach:
Hi @jeremymailen Could I somehow get a build from this PR to be able to easily test it (and compare performance) on my 500k LOC project?

jeremymailen:
> Hi @jeremymailen Could I somehow get a build from this PR to be able to easily test it (and compare performance) on my 500k LOC project?

I usually locally bump the version and `gradle publishToMavenLocal` then you can configure your target project to use maveLocal as a repository and target your local build.

hrach:
Ok, so I've tried this on our 450k LOC project. And the results seem to be the same or even better (originally around 12 secs, this PR times even up to 8 secs). This was done using gradle clean and then gradle lintKotlin runs.

So I've tried more proper benchmarking using gradle-benchmark, but it doesn't clean between runs and ktlint seems to be somehow cached, even manual using --no-build-cache doesn't rerun it properly for unchanged code. Do you have any advice? 

Anyway, will you do some personal benchmarking? Is there anything to make it happen here?

hrach:
Hi @jeremymailen, is there something I can make this happen? Thx

src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
jeremymailen: The reason why I ask is that we include 8 "library" type dependencies from ktlint normally in the plugin and do not include the cli part of ktlint. It would be preferable to keep doing that as to not bring in unnecessary code.

With this method of inclusion, wouldn't we also make the ktlint dependencies in `build.gradle.kts` `compileOnly`?

src/main/kotlin/org/jmailen/gradle/kotlinter/KotlinterPlugin.kt
jeremymailen: Today users can select a different ktlint dependency, overriding the one we depend on by default. Would they be able to use the new `ktlint` configuration to do the same? It seems we hardcode it in this implementation.

jeremymailen (commented):
I was able to test locally on a significant project and performance-wise it appeared to be similar to the old implementation. Also didn't run into any classpath tangles in my environment.

With that fear out of the way, I think the main thing left to figure out is how users would configure the ktlint version since, as it stands, I think this change would take away that feature. If we have a satisfying answer to that, we'd need to rebase and fix the two tests that are failing as well.

hrach:
Closing in favor of #250

PR #252, opened 2022-05-13
find .git repo in parent directory

harrycjy1:
im having trouble to execute installKotlinterPrePushHook in my repository 
because i have .git repo in parent directory than my project placed

it would be nice if kotlinter can find .git repo in parent directories

### DIFF
--- a/src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/GitHookTasks.kt
+++ b/src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/GitHookTasks.kt
@@ -1,8 +1,10 @@
 package org.jmailen.gradle.kotlinter.tasks
 
+import org.eclipse.jgit.lib.Repository
+import org.eclipse.jgit.lib.RepositoryBuilder
 import org.gradle.api.DefaultTask
-import org.gradle.api.provider.Property
-import org.gradle.api.tasks.Input
+import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.tasks.InputDirectory
 import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.TaskAction
 import org.jmailen.gradle.kotlinter.support.VersionProperties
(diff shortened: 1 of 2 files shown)


jeremymailen:
Ok, good thought.

So it's not plausible to have a gradle file at the root of your project and treat it as a multi-module build? I've seen / used this pattern before in polyglot or monorepos and it has some other advantages.

harrycjy1:
thanks for replying

actually i didn't think about multi-module build, in my case some of my repositories does not have gradle file in their root directory and i thought this could be pretty common situation,  that's why i made this PR

but as u say this could be useful to people who have a multi module projects

harrycjy1:
i fix the test error in installHookTaskTest

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/GitHookTasks.kt
```diff
@@ -33,8 +36,18 @@ open class InstallPrePushHookTask : InstallHookTask("pre-push") {
  * Install or update a kotlinter-gradle hook.
  */
 abstract class InstallHookTask(@get:Internal val hookFileName: String) : DefaultTask() {
+
     @Input
-    val gitDirPath: Property<String> = project.objects.property(default = ".git")
+    val gitDirPath = project.objects.property(default = ".git")
+
+    @get:InputDirectory
```
mateuszkwiecinski, line 43: 🐛 ?
`@InputDirectory` documentation states:
>This will cause the task to be considered out-of-date when the directory location **or contents** have changed.

It doesn't sound valid to listen for contents changes of the whole `projectDirectory` - I wouldn't expect the task to be re-run if any file within the project changes 👀

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/GitHookTasks.kt
```diff
@@ -33,8 +36,18 @@ open class InstallPrePushHookTask : InstallHookTask("pre-push") {
...
+    internal val projectDir: DirectoryProperty = project.objects.directoryProperty().apply {
+        set(project.layout.projectDirectory)
+    }
+
+    private val gitRepo: Repository =
+        RepositoryBuilder().findGitDir(projectDir.get().asFile)
```
mateuszkwiecinski, line 49: ❔ 
What's the purpose of `projectDir` property? What would happen if you instead wrote:
```suggestion
    private val gitRepo = RepositoryBuilder().findGitDir(project.projectDir)
```
?

jeremymailen:
Closing some old PRs which are out of date. If we want to approach the feature again, please reopen a new PR.

PR #255, opened 2022-06-18
Draft: feat: register task for pre-commit hook in plugin

bespaltovyj:
Hi, I saw that you have already added a precommit hook, but did not register the task in the plugin. I did this because automatically formatting the code without an error when building a project looks tempting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,8 @@ Granular tasks are added for each source set in the project: `formatKotlin`*`Sou
 
 ### Git Hooks
 
-Kotlinter can install a hook to run pre-push (`installKotlinterPrePushHook`). The hook runs `lintKotlin` and, if there are errors, `formatKotlin` and exits non-zero leaving changed files to be committed.
+* Kotlinter can install a hook to run pre-push (`installKotlinterPrePushHook`). The hook runs `lintKotlin` and, if there are errors, `formatKotlin` and exits non-zero leaving changed files to be committed.
+* Kotlinter can install a hook to run pre-commit (`installKotlinterPreCommitHook`). The hook runs `formatKotlin` and exits non-zero, if auto-formatting failed.
 
 You *must* apply the kotlinter plugin to your root project to make this task available. If using `git worktree` you must install the hook from the parent git directory.
 
@@ -127,6 +128,8 @@ To install the hook automatically when someone runs the build, add this to your 
 ```kotlin
 tasks.check {
     dependsOn("installKotlinterPrePushHook")
+    // or
+    dependsOn("installKotlinterPreCommitHook")
 }
 ```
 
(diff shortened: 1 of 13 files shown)


jeremymailen:
Thanks for the interest @bespaltovyj.
The reason it isn't enabled yet is that probably the logic is too naive. Really it ought to only lint files which are staged for commit and properly action them.

bespaltovyj:
Ha-ha. It's not as easy as it seemed at first blush. 

During the research I encountered the following problems:
1. All files modified in pre-commit hook are not automatically staged to the commit. I was tried manualy add files to index via git-add. It only worked if you did it sequentially, but the project in which I tested it was a multi-module with a parallel build and git index always has broken.
2. It is difficult to understand exactly which files have changed and are included in the commit, because I use `git commit --all` very often myself

So I came to the following conclusion - code formatting via pre-commit hook is evil, because there is a risk of losing your changes. People work is always more important than formatting code, so formatting follows main changes.

I suggest replacing the `pre-commit` hook with a post-commit in which the files from the last commit will be formatted. The `pre commit` hook should be removed.

In my opinion, the work scenario looks like this:
1. The necessary changes are made to the code and they are commited
2. After fixing, formatting starts via `post-commit` hook and the people must commit it himself after viewing.

@jeremymailen @mateuszkwiecinski do you like this solution?

P.S. don't look at the code, because it's raw and it doesn't have what I described.

jeremymailen:
> @jeremymailen @mateuszkwiecinski do you like this solution?
> 
> P.S. don't look at the code, because it's raw and it doesn't have what I described.

Just my own opinion, but as a developer when I want to preemptively auto-format code, I configure my editor to do this so I can see it happen live and examine the results.

The pre-push hook is my preferred solution to saving CI cycles on a run that will fail from lint. Just before I push it will be a safety net to stop the push and show me the lint.

bespaltovyj:
I can offer this option: a post-commit hook that calls the format task and after that there will be an auto commit with code formatting (files will be formatted only from the previous commit).

This option is suitable for those who use git hosting (github, github etc.) and do not push to main remote branches directly. In many systems, you can squash commits when MR(PR) is accepted, so only one commit will remain as a result.

jeremymailen:
> I can offer this option: a post-commit hook that calls the format task and after that there will be an auto commit with code formatting (files will be formatted only from the previous commit).
> 
> This option is suitable for those who use git hosting (github, github etc.) and do not push to main remote branches directly. In many systems, you can squash commits when MR(PR) is accepted, so only one commit will remain as a result.

I appreciate the efforts, but I suggest we close and leave the creation of advanced git hooks up to the end user. It's better I think for them to sequence their git commands in cases where we can't implement an obvious standard workflow in the plugin that's likely to work in all environments.

PR #263, opened 2022-06-27
Bump Kotlin to 1.7.0, bump kotlinter-gradle to 3.11.1

mateuszkwiecinski:
\+ I'm a huge fan of trailing commas, so I toggled related `.editorconfig` properties (which are now recognized by ktlint 🎉)


.editorconfig
```diff
@@ -2,3 +2,6 @@
 max_line_length = 140
 indent_size = 4
 insert_final_newline = true
+ij_kotlin_allow_trailing_comma = true
+ij_kotlin_allow_trailing_comma_on_call_site = true
+disabled_rules=filename
```
mateuszkwiecinski, line 7: If you remove the property, there is just a single offence, claiming I should rename `Utils.kt` to `Resolve.kt`, but I disagree with this rule in general 🤷 If you believe we should adopt all ktlint rules - please let me know, and I'll adjust the code accordingly :)

build.gradle.kts
```diff
@@ -34,6 +34,11 @@ object Versions {
 configurations {
     register("testRuntimeDependencies") {
         extendsFrom(compileOnly.get())
+        attributes {
+            // KGP publishes multiple variants https://kotlinlang.org/docs/whatsnew17.html#support-for-gradle-plugin-variants
+            attribute(Usage.USAGE_ATTRIBUTE, project.objects.named(Usage.JAVA_RUNTIME))
+            attribute(Category.CATEGORY_ATTRIBUTE, project.objects.named(Category.LIBRARY))
+        }
```
mateuszkwiecinski, line 41: This turned out to be pretty neat feature, more info: https://youtrack.jetbrains.com/issue/KT-52966

src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
```diff
@@ -1,8 +1,20 @@
-package org.jmailen.gradle.kotlinter.tasks
+package org.jmailen.gradle.kotlinter.support
 
+import com.pinterest.ktlint.core.api.DefaultEditorConfigProperties
+import com.pinterest.ktlint.core.api.EditorConfigOverride
 import org.gradle.api.file.ProjectLayout
 import java.io.File
 
+internal fun editorConfigOverride(ktLintParams: KtLintParams): EditorConfigOverride {
```
mateuszkwiecinski, line 8: The `KtLintEditorConfigOverrides.kt` was initially reported by `filename` check, so I realized we do already have a _container_ for EditorConfig-related utilities. I moved the utiliy from `.tasks` to `.support` package, as that seemed to be a proper place for such utilities 👀

build.gradle.kts
```diff
@@ -2,11 +2,11 @@ import org.jetbrains.kotlin.gradle.plugin.getKotlinPluginVersion
...
 
 plugins {
-    kotlin("jvm") version "1.6.20"
-    id("com.gradle.plugin-publish") version "0.18.0"
+    kotlin("jvm") version "1.7.0"
+    id("com.gradle.plugin-publish") version "0.21.0"
     `java-gradle-plugin`
     `maven-publish`
-    id("org.jmailen.kotlinter") version "3.10.0"
+    id("org.jmailen.kotlinter") version "3.11.1"
```
jeremymailen, line 9: Should we bundle this with the next ktlint release?

jeremymailen (commented):
Anything in the test GHA test matrix to update?

build.gradle.kts
mateuszkwiecinski: I might be missing something, but that's a plugin's build dependency. It doesn't affect the produced artifact in any way. Running `./gradlew :dependencies --no-configuration-cache | grep kotlinter` returns no kotlinter dependency present.

mateuszkwiecinski:
There was more recent release of Kotlin, but given you've already reviewed most of the changes here - I'll bump it separately.

>Anything in the test GHA test matrix to update?

I checked and all seems to be up-to-date 👍 `integration-tests` tests various `OS`/`gradle`/`java` configurations, `integration-tests-android` switches between various `Gradle`/`Java`/`AGP`. Kotlin bump didn't change any of them, so we should be still okay

.editorconfig
jeremymailen: We can leave it be for now.

jeremymailen (approved):
Good to go!

PR #265, opened 2022-06-27
Reload Ktlint's caches when `.editorconfig` changes

### LINKED ISSUE #262: Regression:  3.11.1 doesn't seem to read `disabled_rules` from .editorconfig anymore
After upgrading from 3.10.0 I had to specify `disabledRules` in the gradle `kotlinter { }` block instead of relying on `disabled_rules` in .editorconfig.

mateuszkwiecinski:
Fixes #262

It seems like this was the scenario I mentioned in: https://github.com/jeremymailen/kotlinter-gradle/pull/246#discussion_r841261149 😅 

My proposal is to do basically the same thing as ktlint-gradle does 🤷 Maaybe if https://github.com/pinterest/ktlint/issues/1446 gets addressed we'll get appropriate api to know when something changes (or ktlint will be smart enough to track `.editorconfig` changes)

### DIFF
--- a/src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
+++ b/src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
@@ -3,9 +3,9 @@ package org.jmailen.gradle.kotlinter.support
 import com.pinterest.ktlint.core.KtLint
 import com.pinterest.ktlint.core.api.DefaultEditorConfigProperties
 import com.pinterest.ktlint.core.api.EditorConfigOverride
+import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.file.ProjectLayout
 import org.gradle.api.logging.Logger
-import org.gradle.api.provider.Property
 import java.io.File
 
 internal fun editorConfigOverride(ktLintParams: KtLintParams): EditorConfigOverride {
@@ -19,12 +19,13 @@ internal fun editorConfigOverride(ktLintParams: KtLintParams): EditorConfigOverr
 }
 
 internal fun resetEditorconfigCacheIfNeeded(
-    wasEditorConfigChanged: Property<Boolean>,
+    changedEditorconfigFiles: ConfigurableFileCollection,
     logger: Logger,
 ) {
-    if (wasEditorConfigChanged.get()) {
+    val changedFiles = changedEditorconfigFiles.files
+    if (changedFiles.any()) {
         logger.info("Editorconfig changed, resetting KtLint caches")
-        KtLint.trimMemory() // Calling trimMemory() will also reset internal loaded `.editorconfig` cache
+        changedFiles.map(File::toPath).forEach(KtLint::reloadEditorConfigFile)
     }
 }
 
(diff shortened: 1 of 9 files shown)


henrik242:
Any progress on this? @jeremymailen @mateuszkwiecinski

jeremymailen:
> Any progress on this? @jeremymailen @mateuszkwiecinski

Sorry for the delay. I was on vacation and then sick, so side projects haven't gotten much attention lately.
I'll try to get comments out on this one and other over the next few days.

src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
```diff
@@ -0,0 +1,55 @@
...
+    wasEditorConfigChanged: Property<Boolean>,
+    logger: Logger,
+) {
+    if (wasEditorConfigChanged.get()) {
+        logger.info("Editorconfig changed, resetting KtLint caches")
+        KtLint.trimMemory() // Calling trimMemory() will also reset internal loaded `.editorconfig` cache
+    }
+}
+
+internal fun ProjectLayout.findApplicableEditorConfigFiles(): Sequence<File> {
```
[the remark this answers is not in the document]
paul-dingemans: I am about to release Ktlint 0.47 when I stumbled on a link to this issue. Please see my remarks in the [ktlint issue](https://github.com/pinterest/ktlint/issues/1446#issuecomment-1217058654). If we can come to a nagreement quickly, I am willing to make an attempt to provide a new API in this 0.47 release.

src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
```diff
@@ -0,0 +1,55 @@
...
+    val projectEditorConfig = projectDirectory.file(".editorconfig").asFile
+
+    return generateSequence(seed = projectEditorConfig) { editorconfig ->
+        if (editorconfig.isRootEditorConfig) {
+            null
+        } else {
+            editorconfig.parentFile?.parentFile?.resolve(".editorconfig")
+        }
+    }
```
paul-dingemans, line 40: This implementation does find `.editorconfig` files located in the current directory or in any parent. Also it is nice that you stop visiting parent directories when you encounter the "root=true" property.
What I do miss however, but I might be wrong, is checking for `.editorconfig` files inside the project itself. Don't they need to be detected and listed beforehand?

jeremymailen:
I assume we'll update this PR after https://github.com/jeremymailen/kotlinter-gradle/pull/275?
Seems good to bundle in a kotlinter 3.12

src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
mateuszkwiecinski: >What I do miss however, but I might be wrong, is checking for .editorconfig files inside the project itself. Don't they need to be detected and listed beforehand?

I think this is a valid comment 👀 Although I'd avoid blindly traversing through project sources looking for potential editorconfig files. I claim most project use single editorconfig defined at Gradle's `rootProject` + in advanced use cases users _override_ its settings by setting another `editorconfig` at Gradle `project` level (i.e. to have different formatting for generated files committed to the repository). 
I'll still suggest going with what I proposed here, at least for now, as I believe it addresses most cases plus most likely it'll be fixed when we switch to the new api you'll introduce for https://github.com/pinterest/ktlint/issues/1446

mateuszkwiecinski:
>I assume we'll update this PR after https://github.com/jeremymailen/kotlinter-gradle/pull/275?

Yes exactly. I waited until `0.47.0` bump is merged so I can switch to the new api, to avoid reloading _all_ caches

jeremymailen (approved):
Excellent!

PR #273, opened 2022-08-19
Use the correct sourceSets

Goooler:
Ref https://github.com/cashapp/sqldelight/pull/3419.


Goooler:
Sorry, seems just no need to correct.

PR #278, opened 2022-09-10
Bump kotlinter 3.12.0 + update integration test matrix

mateuszkwiecinski:
1. Bump `kotlinter-gradle` to recently released `3.12.0`
2. Run android integration tests against dynamically set latest `stable` and `alpha` AGP versions:
3. Seeing changes in #250, improve integration tests verification by failing the workflow if expected ktlint check wasn't verified. My idea here is to verify the custom check was invoked, as expected.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,7 +1,9 @@
+root = true
+
 [*.{kt,kts}]
 max_line_length = 140
 indent_size = 4
 insert_final_newline = true
 ij_kotlin_allow_trailing_comma = true
 ij_kotlin_allow_trailing_comma_on_call_site = true
-disabled_rules=filename
+ktlint_disabled_rules = filename
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -77,14 +77,27 @@ jobs:
           gradle-version: ${{ matrix.gradle }}
           arguments: formatKotlin lintKotlin
 
+  provide-agp-version-matrix:
+    runs-on: ubuntu-latest
+    outputs:
+      agp-versions: ${{ steps.build-agp-matrix.outputs.agp-versions }}
+    steps:
+      - id: agp-version-finder
+        uses: usefulness/agp-version-finder-action@v1
+
+      - id: build-agp-matrix
+        run: echo '::set-output name=agp-versions::["${{ steps.agp-version-finder.outputs.latest-stable }}", "${{ steps.agp-version-finder.outputs.latest-alpha }}"]'
+
   integration-tests-android:
     runs-on: ubuntu-latest
+    needs:
+      - provide-agp-version-matrix
     strategy:
       fail-fast: false
       matrix:
         gradle: [ current ]
         java: [ 17 ]
-        agp: [ 7.0.4 ]
+        agp: ${{ fromJSON(needs.provide-agp-version-matrix.outputs.agp-versions) }}
         include:
           - gradle: 7.0.1
             java: 11
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -77,6 +77,19 @@ jobs:
           gradle-version: ${{ matrix.gradle }}
           arguments: formatKotlin lintKotlin
 
+      - name: Assert fixes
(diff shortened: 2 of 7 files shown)


.editorconfig
```diff
@@ -1,7 +1,9 @@
+root = true
+
 [*.{kt,kts}]
 max_line_length = 140
 indent_size = 4
 insert_final_newline = true
 ij_kotlin_allow_trailing_comma = true
 ij_kotlin_allow_trailing_comma_on_call_site = true
-disabled_rules=filename
+ktlint_disabled_rules = filename
```
mateuszkwiecinski, line 9: 1. `disabled_rules` is now deprecated and shows warnings

.github/workflows/default.yml
```diff
@@ -77,14 +77,40 @@ jobs:
...
     runs-on: ubuntu-latest
+    needs:
+      - provide-agp-version-matrix
     strategy:
       fail-fast: false
       matrix:
         gradle: [ current ]
         java: [ 17 ]
-        agp: [ 7.0.4 ]
+        agp: ${{ fromJSON(needs.provide-agp-version-matrix.outputs.agp-versions) }}
```
mateuszkwiecinski, line 113: the `7.2.2` and `7.4.0-alpha10` are calculated dynamically now

image
image


We'll always test against stable both Gradle and AGP + we'll know early when alpha AGP version breaks things

And also a note: this is my personal action i wrote very recently, for exactly the same purpose (I got annoyed by having manually bumping agp version in CI pipeline)
Related article that inspired me to test early against alpha versions: https://slack.engineering/shadow-jobs/

.github/workflows/default.yml
```diff
@@ -77,14 +77,40 @@ jobs:
           gradle-version: ${{ matrix.gradle }}
           arguments: formatKotlin lintKotlin
 
+      - name: Assert fixes
+        run: |
+          cd test-project/build/reports/ktlint
+          grep 'no-empty-class-body' main-format.txt | grep -q 'EmptyClassBodyClass.kt:3:27'
+          grep 'op-spacing' test-format.txt | grep -q 'OpSpacing.kt:5:16'
```
mateuszkwiecinski, line 84: I wanted to make the pipeline fail when custom rule added in #250 isn't added and called. The assumption here is the `grep 'custom-rule-name' test-format.txt` will find the line in `formatKotlin` task report and will make sure the offence was logged at the right spot:

image

test-project-android/build.gradle.kts
```diff
@@ -6,4 +6,11 @@ plugins {
 
 android {
     compileSdkVersion(31)
+
+    if (properties["agpVersion"].toString().startsWith("4")) {
+        sourceSets {
+            named("main") { java.srcDirs("src/main/kotlin") }
+            named("test") { java.srcDirs("src/test/kotlin") }
+        }
+    }
```
mateuszkwiecinski, line 15: These new checks revealed a bug, `src/xxx/kotlin` wasn't registered as sources on AGP <7.0 by default. Our matrix reported errors only within `src/main/java` directory when invoked with AGP 4.2.2 😬

jeremymailen: Nice fix

.github/workflows/default.yml
jeremymailen: ok

jeremymailen: But let's be mindful of getting too clever. I always ask the question -- will the setup run a year from now? And will someone know how to fix it if it doesn't?

jeremymailen (approved):
Thank you!

.github/workflows/default.yml
mateuszkwiecinski: >will the setup run a year from now?

I hope it will, it should 😅

>And will someone know how to fix it if it doesn't?

I personally am not that afraid of trying such things. If it breaks, one can a) ignore the failing check (it isn't required and won't block the development), b) easily revert the commit, c) simply replace the values with hardcoded ones.

I got your point though 👍

PR #285, opened 2022-09-28
Migrate to JUnit5 + bump AGP to 7.3.0 + bump Kotlin to 1.7.20

mateuszkwiecinski:
In #279, I'll want to temporarily disable some tests onWindows due to Gradle test-kit bug: https://github.com/gradle/gradle/issues/21964 
JUnit5 has a dedicated [DisabledOnOs](https://junit.org/junit5/docs/5.2.0/api/org/junit/jupiter/api/condition/DisabledOnOs.html) annotation, so I'm proposing to upgrate test suite to run with JUnit5, which would be nice to do even if I didn't need the new annotation :)

### DIFF
--- a/src/test/kotlin/org/jmailen/gradle/kotlinter/functional/WithGradleTest.kt
+++ b/src/test/kotlin/org/jmailen/gradle/kotlinter/functional/WithGradleTest.kt
@@ -7,8 +7,6 @@ import org.gradle.testkit.runner.internal.PluginUnderTestMetadataReading
 import org.junit.jupiter.api.io.TempDir
 import java.io.File
 
-val File.root get() = this
-
 abstract class WithGradleTest {
 
     @TempDir
--- a/src/test/kotlin/org/jmailen/gradle/kotlinter/functional/WithGradleTest.kt
+++ b/src/test/kotlin/org/jmailen/gradle/kotlinter/functional/WithGradleTest.kt
@@ -1,16 +1,31 @@
 package org.jmailen.gradle.kotlinter.functional
 
+import org.apache.commons.io.FileUtils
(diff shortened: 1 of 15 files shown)


mateuszkwiecinski:
Ehh... :/ https://github.com/junit-team/junit5/issues/2811

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/WithGradleTest.kt
```diff
@@ -1,16 +1,31 @@
...
+     */
+    @BeforeEach
+    internal fun setUpTempdir() {
+        testProjectDir = Files.createTempDirectory(this::class.java.simpleName).toFile()
+    }
+
+    @AfterEach
+    internal fun cleanUpTempdir() {
+        FileUtils.forceDeleteOnExit(testProjectDir)
+    }
```
mateuszkwiecinski, line 28: Normally, I'd replace `get:Rule` with a `@TempDir` annotation, but unfortunately, it seems like JUnit5 tries to delete temp files at a different time than JUnit4 which doesn't play well with test-kit behavior. I applied one of the workarounds I found in the linked issue

jeremymailen (approved):
I assume no plugin compatibility changes here? User will not need to be on later versions of kotlin or agp?

mateuszkwiecinski:
>I assume no plugin compatibility changes here? User will not need to be on later versions of kotlin or agp?

No changes needed, we should be all good. Kotlin 1.7.20 is backwards compatible and AGP isn't exposed as a public dependency, we'll always use the one provided by the consumer build (that's why I pushed for testing against various AGP versions)

image

Issue #264, opened 2022-06-27
Improve performance by making `LintTask` incremental


mateuszkwiecinski:
Looking at the [source code](https://github.com/jeremymailen/kotlinter-gradle/blob/9931d2965ef076100ea2a0bf1f42dfc66d320172/src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/LintTask.kt#L50), running the `lintKotlin` will result in all the sources being checked on each task re-run. 

To improve the performance, `LintTask` could run against changed files only ([documentation](https://docs.gradle.org/current/userguide/custom_tasks.html#incremental_tasks))

If this gets implemented, a similar thing could be adopted for `formatKotlin` task (with slightly bigger effort due to overlapping inputs and outputs), but would help implementing: https://github.com/jeremymailen/kotlinter-gradle/issues/155

jeremymailen:
Yes, this one is on my wishlist. We're incremental at the sourceSet level, but if a sourceSet is large we might get a boost from only processing changed files in the sourceSet.

PR #289, opened 2022-10-16
Make `LintTask` incremental 

mateuszkwiecinski:
Fixes: https://github.com/jeremymailen/kotlinter-gradle/issues/264


src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/FormatTask.kt
```diff
@@ -39,7 +39,7 @@ open class FormatTask @Inject constructor(
 
         workQueue.submit(FormatWorkerAction::class.java) { p ->
             p.name.set(name)
-            p.files.from(source)
+            p.files.from(getChangedSources(inputChanges))
```
jeremymailen, line 42: I'm not certain we can apply the same incremental logic to formatKotlin as it changes files itself 🤔. I remember running into problems with this at some point when I tried to make format more incremental (at the sourceSet level) than it was.

mateuszkwiecinski: `getChangedSources` returns changed files only when the task is invoked incrementally, which is never the case for `formatKotlin` so the `getChangedSources` will always return whole set of source files. 
But I see how that might be confusing, I'll consider keeping old behavior to be more explicit on the difference between these two tasks types 👍

jeremymailen (approved):
Looks good. Let me know what test cases we need to go through after integration to validate incremental is working as expected.

PR #290, opened 2022-10-18
Remove JCIFS license to fix #287

### LINKED ISSUE #287: JCIFS referenced in LICENSE
I found the [LGPL included in LICENSE for JCIFS](https://github.com/jeremymailen/kotlinter-gradle/blob/9f200573e54cde1fb0e55b06fa03f85ca27a1630/LICENSE#L458), but I didn't find any reference to JCIFS in the source code. Is JCIFS actually used?

jeremymailen:
We don't use it, not sure why it is in there.


jeremymailen:
@mateuszkwiecinski maybe we don't want the AGP test to auto-upgrade to the latest? We're bound to get failures at some point on a PR which does not change the support but just has bad timing luck.

mateuszkwiecinski (approved):

mateuszkwiecinski:
I personally appreciate seeing it failing eagerly, which lets us know something will break, before it reaches stable channel and does actually have impact on plugin consumers.
I already fixed it in one of my branches (I can pull out the fix to a separate PR).

I would never make it a required check, so it doesn't block the pipeline, but yeah, knowing how unstable AGP is, I still see a value in keeping the check. Issues reported to alpha version have a chance of being fixed, but if they land in a stable channel it's hard to predict if they ever get fixed (example: https://issuetracker.google.com/issues/170650362) 

but you're the _boss_ here, and I'll do as you wish 😛

jeremymailen:
> I personally appreciate seeing it failing eagerly, which lets us know something will break, before it reaches stable channel and does actually have impact on plugin consumers. I already fixed it in one of my branches (I can pull out the fix to a separate PR).
> 
> I would never make it a required check, so it doesn't block the pipeline, but yeah, knowing how unstable AGP is, I still see a value in keeping the check. Issues reported to alpha version have a chance of being fixed, but if they land in a stable channel it's hard to predict if they ever get fixed (example: https://issuetracker.google.com/issues/170650362)
> 
> but you're the _boss_ here, and I'll do as you wish 😛

Seems problematic if someone sees a failure on their PR which is not their fault.

Maybe it makes sense as non-blocking on the main branch workflow only?

mateuszkwiecinski:
>Seems problematic if someone sees a failure on their PR which is not their fault.

Fair point! I'll update CI config to run only on the main branch then 👍

PR #293, opened 2022-10-22
Run workflows against alpha AGP version only on the main branch + fix android test project 

mateuszkwiecinski:
As discussed in: https://github.com/jeremymailen/kotlinter-gradle/pull/290#issuecomment-1283038262

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -99,7 +99,12 @@ jobs:
         uses: usefulness/agp-version-finder-action@v1
 
       - id: build-agp-matrix
-        run: echo '::set-output name=agp-versions::["${{ steps.agp-version-finder.outputs.latest-stable }}", "${{ steps.agp-version-finder.outputs.latest-alpha }}"]'
+        run: |
+           if [[ ${{ github.event_name }} == 'pull_request' ]]; then
+             echo '::set-output name=agp-versions::["${{ steps.agp-version-finder.outputs.latest-stable }}"]'
+           else
+             echo '::set-output name=agp-versions::["${{ steps.agp-version-finder.outputs.latest-stable }}", "${{ steps.agp-version-finder.outputs.latest-alpha }}"]'
+           fi
 
   integration-tests-android:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/default.yml
```diff
@@ -99,7 +99,12 @@ jobs:
         uses: usefulness/agp-version-finder-action@v1
 
       - id: build-agp-matrix
-        run: echo '::set-output name=agp-versions::["${{ steps.agp-version-finder.outputs.latest-stable }}", "${{ steps.agp-version-finder.outputs.latest-alpha }}"]'
+        run: |
+           if [[ ${{ github.event_name }} == 'pull_request' ]]; then
+             echo 'agp-versions=["${{ steps.agp-version-finder.outputs.latest-stable }}"]' >> $GITHUB_OUTPUT
+           else
+             echo 'agp-versions=["${{ steps.agp-version-finder.outputs.latest-stable }}", "${{ steps.agp-version-finder.outputs.latest-alpha }}"]' >> $GITHUB_OUTPUT
+           fi
```
mateuszkwiecinski, line 107: I considered having a separate workflow for each scenario (build main branch vs build pull requests), but decided it's not worth separating yet

.github/workflows/default.yml
```diff
@@ -115,9 +120,6 @@ jobs:
           - gradle: 7.0.1
             java: 11
             agp: 4.2.2
-          - gradle: 6.9.1
-            java: 8
-            agp: 4.0.0
```
mateuszkwiecinski, line 120: AGP introduced `namespace` property, which is available since 4.2 and is required starting from 8.0.0. Given the above, this part of the matrix failed, as the `namespace` I'm adding below isn't supported 👀 
I didn't want to make our configuration too complicated, so I'm proposing to stop testing on that old version 🤷 It doesn't change plugin compatibility, but given we'll stop testing on it - it's not guaranteed it _will_ work.

.github/workflows/default.yml
jeremymailen: Yeah, this seems good enough.

jeremymailen (approved):

mateuszkwiecinski:
I can confirm it all worked as expected: https://github.com/jeremymailen/kotlinter-gradle/actions/runs/3304361928/jobs/5453652284

The readme badge is back to green:
image

🎉

Issue #171, opened 2021-01-10
Implement missing properties in the configuration closure


rock3r:
Right now there is no way to set the `--debug`, `--baseline`, and `--editorconfig` ktlint flags. it would be great to have that possibility. I'd imagine those being additional properties on the `kotlinter` closure:

```kotlin
kotlinter {
    // ...
    debug = false
    baseline = file(...)
    editorconfig = file(...)
}
```

jeremymailen:
Thanks for the note!
The baseline would be a new feature, see #167. Probably based on demand -- do you use it?

Wondering about editorconfig as well -- since it's a build script would it make sense to refer to a personal location over an editorconfig in the standard location checked into the project? Certainly could support it if people want to use it vs allowing ktlint to automatically locate editorconfig in the standard places.

For debug, would it make sense to pass through gradle's --debug flag when it is on? This would align to the same way you'd debug the kotlinter plugin itself.

rock3r:
Hey there!

Baseline: not yet, but I would if it were available, at least in one project I work on.

Editorconfig: would use it mostly for debugging my own editorconfig when the build fails and I can't figure out why :) 

As for debug: tying it with Gradle's by default sounds reasonable, however I'd most likely want to only debug ktlint runs on their own, and their debug output would risk getting lost in Gradle's very, very verbose debug output, I think

jeremymailen:
Ok, makes sense. Will try to put these in queue for next dot release.

rock3r:
Hi there — once again, I'd need the `editorconfig` property. The new use case is that I am applying the Kotlinter configuration to subprojects via a compiled script plugin and it's not picking up the `.editorconfig` file in the root of the repository, so I need to explicitly point to it.

mateuszkwiecinski:
Hey @rock3r 👋  - if `ktlint` doesn't pick your `.editorconfig` you might be observing https://github.com/jeremymailen/kotlinter-gradle/issues/262 and the temporary workaround is to kill gradle daemon.

rock3r:
I'll check and let you know :) Thanks for the pointer

rock3r:
That did the trick, thanks @mateuszkwiecinski 👍

chrisjenx:
THanks @mateuszkwiecinski that fixed a different issue I was seeing

mateuszkwiecinski:
[...]

PR #294, opened 2022-10-22
Load `Reporters` dynamically, using provided `ktlint` version

mateuszkwiecinski:
Follow up on https://github.com/jeremymailen/kotlinter-gradle/pull/279#discussion_r996320177 - which moves all `ktlint` classes to be loaded in process isolation, within workers. The same way `RuleSetProviderV2`s are.


src/main/kotlin/org/jmailen/gradle/kotlinter/support/Reporters.kt
```diff
@@ -1,43 +1,60 @@
...
+    ReporterType.Plain,
+    -> output.toRelativeString(base = relativeRoot)
+
+    ReporterType.Sarif -> output.absolutePath
 }
-/* ktlint-enable enum-entry-name-case */
-
-fun reporterFor(reporterName: String, output: File): Reporter {
-    val out = PrintStream(output)
-    return SortedThreadSafeReporterWrapper(
```
mateuszkwiecinski, line 24: I removed `SortedThreadSafeReporterWrapper` as it doesn't seem to be needed. It was added when the source files was chunked, which isn't the case anymore 👀 

I'd have to go through repo history to remind myself why chunking got dropped, but given that still sounds like a good idea to me, if we considered restoring it I'd re-visit the approach with `SortedThreadSafeReporterWrapper` and rather create a separate task that would do the reporting (to ensure correct output and avoid parallelization issues). 
From my perspective, we're now changing _correctness_ and the focus on performance could be the next step.

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/ReportersTest.kt
```diff
@@ -0,0 +1,216 @@
...
+                }
+
+                """
+            appendText(buildScript)
+        }
+
+        build("lintKotlin").apply {
+            val reportContent = projectRoot.resolve("build/reports/ktlint/main-lint.sarif.json").readText()
+            assertTrue(reportContent.contains(""""version": "0.47.0""""))
+            assertTrue(reportContent.contains(""""semanticVersion": "0.47.0""""))
```
mateuszkwiecinski, line 119: I noticed ktlint version appears in generated report so I decided to use to verify the behavior changes when the `ktlintVersion` gets overriden

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/ReportersTest.kt
```diff
@@ -0,0 +1,216 @@
...
+""".trimIndent()
+
+// language=html
+private fun expectedEmptyHtml() = """
+<html>
+<head>
+<link href="https://fonts.googleapis.com/css?family=Source+Code+Pro" rel="stylesheet" />
+<meta http-equiv="Content-Type" Content="text/html; Charset=UTF-8">
+<style>
+body {
```
mateuszkwiecinski, line 143: Asserting on file content isn't necessarily the most reliable way to ensure proper behavior, and we probably should only compare smaller parts of the content. I wanted to avoid making _decisions_ what to test here specifically, and decided to start with full-content comparison and then migrate to either smaller chunks or start parsing the reports? Ideas are welcome

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/ReportersTest.kt
```diff
@@ -0,0 +1,216 @@
...
+            appendText(buildScript)
+        }
+
+        build("lintKotlin").apply {
+            assertEquals(TaskOutcome.SUCCESS, task(":lintKotlin")?.outcome)
+            assertEquals(expectedEmptyPlain(), reportContent("/main-lint.txt"))
+            assertEquals(expectedEmptyCheckstyle(), reportContent("main-lint.xml"))
+            assertEquals(expectedEmptyHtml(), reportContent("/main-lint.html"))
+            assertEquals(expectedEmptyJson(), reportContent("/main-lint.json"))
+            assertTrue(reportContent("/main-lint.sarif.json").isNotBlank())
```
mateuszkwiecinski, line 73: `sarif` report contains absolute paths which varies between devices. I haven't found any better way to ensure correct report gets generated

src/main/kotlin/org/jmailen/gradle/kotlinter/support/Reporters.kt
```diff
@@ -1,43 +1,60 @@
...
-        else -> output.toRelativeString(projectDir)
-    }
+private fun ReporterType.generateOpt() = when (this) {
+    ReporterType.Checkstyle,
+    ReporterType.Html,
+    ReporterType.Json,
+    ReporterType.Sarif,
+    -> emptyMap()
+
+    ReporterType.Plain -> mapOf("color_name" to "DARK_GRAY")
```
mateuszkwiecinski, line 56: Ah, here's _interesting_ thing: previously we instantiated reporter directly using its constructor: [source](https://github.com/pinterest/ktlint/blob/2ea18f057bfac03313b4e687302d7148f2eeb5cc/ktlint-reporter-plain/src/main/kotlin/com/pinterest/ktlint/reporter/plain/PlainReporter.kt#L16)
but the ReporterProvider has different defaults: [source](https://github.com/pinterest/ktlint/blob/2ea18f057bfac03313b4e687302d7148f2eeb5cc/ktlint-reporter-plain/src/main/kotlin/com/pinterest/ktlint/reporter/plain/PlainReporterProvider.kt#L16)
I'll report that upstream, but for now I decided to preserve the old behavior. if we wanted we could expose the reporterOptions via extensions and let plugin consumers provide their own options. This might be also related to: https://github.com/jeremymailen/kotlinter-gradle/issues/171 (to pass the `verbose` parameter)

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/ReportersTest.kt
jeremymailen: Yeah, that is kind of annoying that Sarif doesn't keep to the convention of the others, but nonetheless.

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/ReportersTest.kt
jeremymailen: Yeah, seems a little heavy. If we just check the content for a small known snippet, what's the real chance we'd have a false positive test pass?

jeremymailen (approved):
Looks good. Let's get this one merged into the bigger change along with #289 and we get that big piece landed for 4.0.0.

jeremymailen:
I believe this is also now solved for in 5.0.0

PR #295, opened 2022-10-22
Fix wrong `apache-commons` dependency + fix CI deprecation warnings

mateuszkwiecinski:
Inspired by the number of warnings in: https://github.com/jeremymailen/kotlinter-gradle/actions/runs/3304708014 - this is my attempt to address them

image

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -15,10 +15,10 @@ jobs:
         os: [ ubuntu-latest, windows-latest ]
     name: '[${{ matrix.os }}] build plugin'
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Set up Java
-        uses: actions/setup-java@v2
+        uses: actions/setup-java@v3
         with:
           distribution: 'adopt'
           java-version: 17
@@ -38,7 +38,7 @@ jobs:
 
       - name: Upload test results
         if: ${{ always() }}
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
         with:
           name: test-results-${{ matrix.os }}
           path: "${{ github.workspace }}/**/build/reports/tests"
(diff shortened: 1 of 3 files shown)


mateuszkwiecinski:
No warnings visible anymore: https://github.com/jeremymailen/kotlinter-gradle/actions/runs/3304813961

.github/workflows/default.yml
```diff
@@ -15,16 +15,16 @@ jobs:
...
       - name: Set up Java
-        uses: actions/setup-java@v2
+        uses: actions/setup-java@v3
         with:
           distribution: 'adopt'
           java-version: 17
 
       - name: Gradle wrapper validation
-        uses: gradle/wrapper-validation-action@v1
+        uses: gradle/wrapper-validation-action@859c33240bd026ce8d5f711f5adcc65c2f8eafc1 # https://github.com/gradle/wrapper-validation-action/issues/66
```
jeremymailen, line 27: Would this be fixed by advancing to node 16 for actions?

jeremymailen (approved):

.github/workflows/default.yml
mateuszkwiecinski: AFAIU we (Github Actions users) don't have a control over the Node version used by Github Actions. Gtihub has [announced](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/) they'll switch the runtime at some point, and most Actions migrated already. 
If the Github advanced node version to 16, the `v1` would simply break our pipeline. The [commit](https://github.com/gradle/wrapper-validation-action/commit/859c33240bd026ce8d5f711f5adcc65c2f8eafc1) I linked here properly supports Node 16.

We could wait until they move the `v1` tag, but I wanted to remove all deprecation warnings already 👀 
I'll keep monitoring their repository and if `v1` will contain necessary changes I'll revert it back, for consistency

PR #302, opened 2022-12-04
Add baseline support

### LINKED ISSUE #167: [Question] How to add baseline to check just like ktlint does?
Ktlint https://github.com/pinterest/ktlint
> to add a baseline to check against prepend following args: "--baseline=ktlint-baseline.xml"

dallasgutauckis:
Resolves #167

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,14 +101,14 @@ plugins {
 - Incremental build support and fast parallelization with Gradle Worker API
 - Configures from `.editorconfig` when available
 - Configurable reporters
+- Baseline file _(note: you must generate your baseline file with [`ktlint`](https://github.com/pinterest/ktlint))_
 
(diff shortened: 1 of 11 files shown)


rs-georg:
Would be great if this PR could be merged and released.
Not having baseline support is keeping us from using this plugin. We want to migrate from https://github.com/JLLeitschuh/ktlint-gradle. However, we have existing baseline files that we cannot drop since fixing all the issues would create an unacceptable effort.

ThomasRichtsfeld:
We are also currently integrating this plugin but need baseline support as `formatKotlin` doesn't resolve all the issues. 
Any help needed

jeremymailen:
Closing for now since it's out of date. We can recreate if we plan to add baseline support.

PR #304, opened 2022-12-16
Bump ktlint to 0.48.1

lwasyl:
I'm not actually sure this works, I just wondered how complex the update would be

### DIFF
--- a/src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
+++ b/src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
@@ -1,31 +1,75 @@
 package org.jmailen.gradle.kotlinter.support
 
-import com.pinterest.ktlint.core.KtLint
-import com.pinterest.ktlint.core.api.DefaultEditorConfigProperties
+import com.pinterest.ktlint.core.KtLintRuleEngine
 import com.pinterest.ktlint.core.api.EditorConfigOverride
+import com.pinterest.ktlint.core.api.editorconfig.EditorConfigProperty
+import org.ec4j.core.model.PropertyType
+import org.ec4j.core.model.PropertyType.PropertyValueParser.IDENTITY_VALUE_PARSER
 import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.file.ProjectLayout
 import org.gradle.api.logging.Logger
 import java.io.File
 
-internal fun editorConfigOverride(ktLintParams: KtLintParams): EditorConfigOverride {
-    val rules = ktLintParams.disabledRules
+internal fun editorConfigOverride(ktLintParams: KtLintParams) =
+    getPropertiesForDisabledRules(ktLintParams)
+        .plus(getPropertiesForExperimentalRules(ktLintParams))
+        .let(::buildEditorConfigOverride)
 
+private fun getPropertiesForDisabledRules(
+    ktLintParams: KtLintParams,
+): List<Pair<EditorConfigProperty<String>, String>> {
+    val rules = ktLintParams.disabledRules
     return if (rules.isEmpty()) {
-        EditorConfigOverride.emptyEditorConfigOverride
+        emptyList()
     } else {
-        EditorConfigOverride.from(DefaultEditorConfigProperties.ktlintDisabledRulesProperty to rules.joinToString(separator = ","))
+        rules
+            .asSequence()
+            .map(::getKtlintRulePropertyName)
+            .map { propertyName ->
+                EditorConfigProperty(
+                    type = PropertyType(propertyName, "Rule to be disabled", IDENTITY_VALUE_PARSER),
+                    defaultValue = "disabled",
+                )
+            }
+            .map { it to "disabled" }
+            .toList()
     }
 }
 
-internal fun resetEditorconfigCacheIfNeeded(
+private fun getPropertiesForExperimentalRules(ktLintParams: KtLintParams) =
+    if (ktLintParams.experimentalRules) {
+        listOf(
+            EditorConfigProperty(
+                type = PropertyType("ktlint_experimental", "Experimental rules", IDENTITY_VALUE_PARSER),
+                defaultValue = "enabled",
+            ) to "enabled",
+        )
+    } else {
+        emptyList()
+    }
+
+private fun buildEditorConfigOverride(editorConfigProperties: List<Pair<EditorConfigProperty<String>, String>>) =
+    if (editorConfigProperties.isEmpty()) {
+        EditorConfigOverride.EMPTY_EDITOR_CONFIG_OVERRIDE
+    } else {
+        EditorConfigOverride.from(*editorConfigProperties.toTypedArray())
+    }
+
+private fun getKtlintRulePropertyName(ruleName: String) =
+    if (ruleName.contains(':')) { // Rule from a non-standard rule set
+        "ktlint_${ruleName.replace(':', '_')}"
+    } else {
+        "ktlint_standard_$ruleName"
+    }
+
+internal fun KtLintRuleEngine.resetEditorconfigCacheIfNeeded(
     changedEditorconfigFiles: ConfigurableFileCollection,
     logger: Logger,
 ) {
     val changedFiles = changedEditorconfigFiles.files
     if (changedFiles.any()) {
         logger.info("Editorconfig changed, resetting KtLint caches")
-        changedFiles.map(File::toPath).forEach(KtLint::reloadEditorConfigFile)
+        changedFiles.map(File::toPath).forEach(::reloadEditorConfigFile)
     }
 }
 
--- a/src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
+++ b/src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
@@ -1,19 +1,24 @@
 package org.jmailen.gradle.kotlinter.support
 
-import com.pinterest.ktlint.core.KtLintRuleEngine
+import com.pinterest.ktlint.core.api.EditorConfigDefaults
 import com.pinterest.ktlint.core.api.EditorConfigOverride
 import com.pinterest.ktlint.core.api.editorconfig.EditorConfigProperty
+import org.ec4j.core.model.EditorConfig
+import org.ec4j.core.model.Glob
+import org.ec4j.core.model.Property
 import org.ec4j.core.model.PropertyType
 import org.ec4j.core.model.PropertyType.PropertyValueParser.IDENTITY_VALUE_PARSER
-import org.gradle.api.file.ConfigurableFileCollection
+import org.ec4j.core.model.Section
 import org.gradle.api.file.ProjectLayout
-import org.gradle.api.logging.Logger
 import java.io.File
 
 internal fun editorConfigOverride(ktLintParams: KtLintParams) =
     getPropertiesForDisabledRules(ktLintParams)
-        .plus(getPropertiesForExperimentalRules(ktLintParams))
         .let(::buildEditorConfigOverride)
+
+internal fun editorConfigDefaults(ktLintParams: KtLintParams): EditorConfigDefaults =
+    getPropertiesForExperimentalRules(ktLintParams)
+        .let(::buildEditorConfigDefaults)
 
 private fun getPropertiesForDisabledRules(
     ktLintParams: KtLintParams,
@@ -37,16 +42,11 @@ private fun getPropertiesForDisabledRules(
 }
 
 private fun getPropertiesForExperimentalRules(ktLintParams: KtLintParams) =
-    if (ktLintParams.experimentalRules) {
-        listOf(
-            EditorConfigProperty(
-                type = PropertyType("ktlint_experimental", "Experimental rules", IDENTITY_VALUE_PARSER),
-                defaultValue = "enabled",
-            ) to "enabled",
-        )
-    } else {
-        emptyList()
-    }
+    Property
+        .builder()
+        .name("ktlint_experimental")
+        .value(if (ktLintParams.experimentalRules) "enabled" else "disabled")
+        .let(::listOf)
 
 private fun buildEditorConfigOverride(editorConfigProperties: List<Pair<EditorConfigProperty<String>, String>>) =
     if (editorConfigProperties.isEmpty()) {
@@ -55,23 +55,29 @@ private fun buildEditorConfigOverride(editorConfigProperties: List<Pair<EditorCo
         EditorConfigOverride.from(*editorConfigProperties.toTypedArray())
     }
 
+private fun buildEditorConfigDefaults(kotlinSectionProperties: List<Property.Builder>) =
+    if (kotlinSectionProperties.isEmpty()) {
+        EditorConfigDefaults.EMPTY_EDITOR_CONFIG_DEFAULTS
+    } else {
+        EditorConfigDefaults(
+            EditorConfig
+                .builder()
+                .section(
+                    Section
+                        .builder()
+                        .glob(Glob("*.{kt,kts}"))
+                        .properties(kotlinSectionProperties),
+                )
+                .build(),
+        )
+    }
+
 private fun getKtlintRulePropertyName(ruleName: String) =
     if (ruleName.contains(':')) { // Rule from a non-standard rule set
         "ktlint_${ruleName.replace(':', '_')}"
     } else {
         "ktlint_standard_$ruleName"
     }
-
-internal fun KtLintRuleEngine.resetEditorconfigCacheIfNeeded(
-    changedEditorconfigFiles: ConfigurableFileCollection,
-    logger: Logger,
-) {
-    val changedFiles = changedEditorconfigFiles.files
-    if (changedFiles.any()) {
-        logger.info("Editorconfig changed, resetting KtLint caches")
-        changedFiles.map(File::toPath).forEach(::reloadEditorConfigFile)
-    }
-}
 
 internal fun ProjectLayout.findApplicableEditorConfigFiles(): Sequence<File> {
     val projectEditorConfig = projectDirectory.file(".editorconfig").asFile
--- a/src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/LintTask.kt
+++ b/src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/LintTask.kt
@@ -52,7 +52,7 @@ open class LintTask @Inject constructor(
                 p.projectDirectory.set(projectLayout.projectDirectory.asFile)
                 p.reporters.putAll(reports)
                 p.ktLintParams.set(getKtLintParams())
-                p.changedEditorconfigFiles.from(getChangedEditorconfigFiles(inputChanges))
+                p.changedEditorConfigFiles.from(getChangedEditorconfigFiles(inputChanges))
             }
             runCatching { await() }
         }
--- a/src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/format/FormatWorkerAction.kt
+++ b/src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/format/FormatWorkerAction.kt
@@ -1,7 +1,6 @@
 package org.jmailen.gradle.kotlinter.tasks.format
 
 import com.pinterest.ktlint.core.Code
-import com.pinterest.ktlint.core.KtLintRuleEngine
 import org.gradle.api.logging.LogLevel
 import org.gradle.api.logging.Logger
 import org.gradle.api.logging.Logging
@@ -9,10 +8,8 @@ import org.gradle.internal.logging.slf4j.DefaultContextAwareTaskLogger
 import org.gradle.workers.WorkAction
 import org.jmailen.gradle.kotlinter.support.KotlinterError
 import org.jmailen.gradle.kotlinter.support.KtLintParams
-import org.jmailen.gradle.kotlinter.support.defaultRuleSetProviders
-import org.jmailen.gradle.kotlinter.support.editorConfigOverride
+import org.jmailen.gradle.kotlinter.support.createKtlintEngine
 import org.jmailen.gradle.kotlinter.support.resetEditorconfigCacheIfNeeded
-import org.jmailen.gradle.kotlinter.support.resolveRuleProviders
 import org.jmailen.gradle.kotlinter.tasks.FormatTask
 import java.io.File
 
@@ -25,11 +22,7 @@ abstract class FormatWorkerAction : WorkAction<FormatWorkerParameters> {
     private val output: File? = parameters.output.asFile.orNull
 
     override fun execute() {
-        val ktLintEngine = KtLintRuleEngine(
-            ruleProviders = resolveRuleProviders(defaultRuleSetProviders),
-            editorConfigOverride = editorConfigOverride(ktLintParams),
-        )
-
+        val ktLintEngine = createKtlintEngine(ktLintParams = ktLintParams)
         ktLintEngine.resetEditorconfigCacheIfNeeded(
             changedEditorconfigFiles = parameters.changedEditorConfigFiles,
             logger = logger,
--- a/src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/format/FormatWorkerAction.kt
+++ b/src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/format/FormatWorkerAction.kt
@@ -26,7 +26,7 @@ abstract class FormatWorkerAction : WorkAction<FormatWorkerParameters> {
 
     override fun execute() {
         val ktLintEngine = KtLintRuleEngine(
-            ruleProviders = resolveRuleProviders(defaultRuleSetProviders, ktLintParams.experimentalRules),
+            ruleProviders = resolveRuleProviders(defaultRuleSetProviders),
             editorConfigOverride = editorConfigOverride(ktLintParams),
         )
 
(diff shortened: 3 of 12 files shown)


src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
```diff
@@ -12,20 +15,39 @@ internal fun editorConfigOverride(ktLintParams: KtLintParams): EditorConfigOverr
...
     }
 }
 
-internal fun resetEditorconfigCacheIfNeeded(
+private fun getDisabledRulePropertyName(it: String) =
+    if (it.contains(':')) { // Rule from a custom source set
+        "ktlint_${it.replace(':', '_')}"
+    } else {
+        "ktlint_standard_$it"
+    }
```
lwasyl, line 41: Each disabled rule has to have `ktlint_$ruleset` prepended. From what I see, Kotlinter assumes `standard` default ruleset, I tried to keep that behavior

src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
```diff
@@ -12,20 +15,39 @@ internal fun editorConfigOverride(ktLintParams: KtLintParams): EditorConfigOverr
...
+
+internal fun KtLintRuleEngine.resetEditorconfigCacheIfNeeded(
```
lwasyl, line 43: `KtLint` is now deprecated in favor of `KtLintRuleEngine` instance, as far as I understand. I pulled it as a receiver where needed

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/format/FormatWorkerAction.kt
```diff
@@ -26,29 +25,33 @@ abstract class FormatWorkerAction : WorkAction<FormatWorkerParameters> {
...
+
+        ktLintEngine.resetEditorconfigCacheIfNeeded(
             changedEditorconfigFiles = parameters.changedEditorConfigFiles,
             logger = logger,
         )
 
         val fixes = mutableListOf<String>()
         try {
             files.forEach { file ->
-                val ruleSets = resolveRuleProviders(defaultRuleSetProviders, ktLintParams.experimentalRules)
```
lwasyl, line 37: I wasn't sure why rulesets were resolved for each file, was there a reason for that? Did I break something by moving rules resolution out of the loop?

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/format/FormatWorkerAction.kt
```diff
@@ -73,27 +76,19 @@ abstract class FormatWorkerAction : WorkAction<FormatWorkerParameters> {
             },
         )
     }
+}
 
-    private fun formatKt(file: File, ruleSets: Set<RuleProvider>, onError: ErrorHandler) =
-        format(file, ruleSets, onError, false)
+private fun KtLintRuleEngine.formatKt(file: File, onError: ErrorHandler) =
```
lwasyl, line 81: Moved out of the file to keep function references working after I added a receiver

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/format/FormatWorkerAction.kt
```diff
@@ -73,27 +76,19 @@ abstract class FormatWorkerAction : WorkAction<FormatWorkerParameters> {
...
+private fun KtLintRuleEngine.formatKts(file: File, onError: ErrorHandler) =
+    format(file, onError)
 
-    private fun format(file: File, ruleProviders: Set<RuleProvider>, onError: ErrorHandler, script: Boolean): String {
-        return KtLint.format(
-            KtLint.ExperimentalParams(
-                fileName = file.path,
-                text = file.readText(),
-                ruleProviders = ruleProviders,
-                script = script,
```
lwasyl, line 89: Looks like ktlint now checks if the file is a script internally:
image

I missed that initially, looks like I can remove the kt/kts distinction from this file?

src/test/kotlin/org/jmailen/gradle/kotlinter/functional/CustomTaskTest.kt
```diff
@@ -52,6 +52,7 @@ class CustomTaskTest : WithGradleTest.Kotlin() {
                         println("go")
                     }
                 }
+
```
lwasyl, line 55: A new (I think? haven't looked into it) rule failed with `no newline at end of file`.  Perhaps that was a valid test and the rule is disabled somewhere? Will take a look

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/LintTask.kt
```diff
@@ -52,7 +52,7 @@ open class LintTask @Inject constructor(
                 p.projectDirectory.set(projectLayout.projectDirectory.asFile)
                 p.reporters.putAll(reports)
                 p.ktLintParams.set(getKtLintParams())
-                p.changedEditorconfigFiles.from(getChangedEditorconfigFiles(inputChanges))
+                p.changedEditorConfigFiles.from(getChangedEditorconfigFiles(inputChanges))
```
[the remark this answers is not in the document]
lwasyl: Now I see that there are both variants used. I'll keep this since `FormatTask` uses `EditorConfig` variant

jeremymailen:
Very much appreciate you working up the changes for the 0.48.0 update. I'll try to find some time over the next week or two to test and review. Thanks in advance for your patience -- some holiday travel coming up and work is extraordinarily hectic as well.

I've also noticed that ktlint intends to do a patch release soon to fix some regressions so we probably will want to jump to that version.

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/format/FormatWorkerAction.kt
mateuszkwiecinski: >I wasn't sure why rulesets were resolved for each file, was there a reason for that?

I'm not aware of any 🤷 I agree the change you proposed should work just fine 👀

src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
```diff
@@ -12,20 +15,39 @@ internal fun editorConfigOverride(ktLintParams: KtLintParams): EditorConfigOverr
...
+            }
+            .map { it to "disabled" }
+            .toList()
+            .toTypedArray()
+            .let { EditorConfigOverride.from(*it) }
     }
 }
 
-internal fun resetEditorconfigCacheIfNeeded(
+private fun getDisabledRulePropertyName(it: String) =
```
mateuszkwiecinski, line 36: 🎨 
```suggestion
private fun getDisabledRulePropertyName(ruleName: String) =
```

mateuszkwiecinski: \+ I can propose dropping the `disabled` suffix and clarify this method transform rule name to Ktlint EditorConfig property name
```suggestion
private fun getKtlintPropertyName(ruleName: String) =
```

src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
```diff
@@ -12,20 +15,39 @@ internal fun editorConfigOverride(ktLintParams: KtLintParams): EditorConfigOverr
...
+    if (it.contains(':')) { // Rule from a custom source set
```
mateuszkwiecinski, line 37: ⛏️ 
```suggestion
    if (it.contains(':')) { // Rule from a custom rule set
```

📝 
I'll comment out this also includes rules from experimental rule set so maybe:
```suggestion
    if (it.contains(':')) { // Rule from a non-standard ruleset
```

src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
```diff
@@ -12,20 +15,39 @@ internal fun editorConfigOverride(ktLintParams: KtLintParams): EditorConfigOverr
     val rules = ktLintParams.disabledRules
 
     return if (rules.isEmpty()) {
-        EditorConfigOverride.emptyEditorConfigOverride
+        EMPTY_EDITOR_CONFIG_OVERRIDE
     } else {
-        EditorConfigOverride.from(DefaultEditorConfigProperties.ktlintDisabledRulesProperty to rules.joinToString(separator = ","))
+        rules
```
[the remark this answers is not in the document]
mateuszkwiecinski: 🔧 
My only suggestion here would be to add additional set of functional tests to ensure most common scenarios 
i.e. experimental rules can also be disabled. I don't think we have that covered anywhere 👀

mateuszkwiecinski (commented):
Couple of minor comments, but otherwise LGTM 👍

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/LintTask.kt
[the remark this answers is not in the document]
lwasyl: I think I just copied this call from `FormatTask` and it used different capitalization, so it made sense to unify it. I can revert back to `Editorconfig` here if you prefer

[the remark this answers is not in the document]
mateuszkwiecinski: Actually, I was wrong and you were right. The method I referenced is our own extension, which follows the same _wrong_ convention. I'll unify all names to keep them _consistent_ within the project. Thanks for spotting that 👍

src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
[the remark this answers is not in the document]
lwasyl: Well, good catch: I added a test for disabling an experimental rule (which is not in the _standard_ ruleset) and it passed, but at the same time I added a positive test case to check that the experimental rule is executed in the first place if it's not disabled — it's not. This seems to be an unintended [behavior change](https://github.com/pinterest/ktlint/issues/1747) in ktlint 0.48.0, I assume it's fixed in 0.48.1. For now the PR will fail tests, hopefully after updating ktlint dependency to 0.48.1 it will pass.

As for using experimental rules: I understand it's not ideal and the tests would have to change as experimental rules evolve, but I didn't see an easy way to use a test-only custom rule set in the functional tests. If you have a suggestion on how to do that, I can try

.editorconfig
```diff
@@ -6,4 +6,4 @@ indent_size = 4
 insert_final_newline = true
 ij_kotlin_allow_trailing_comma = true
 ij_kotlin_allow_trailing_comma_on_call_site = true
-ktlint_disabled_rules = filename
+ktlint_standard_filename = disabled
```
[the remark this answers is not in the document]
jeremymailen: Yeah, the project itself is briefly behind the latest version, which is hard to avoid.

src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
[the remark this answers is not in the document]
jeremymailen: 0.48.1 is out now, so we can bump this PR to that perhaps?

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/format/FormatWorkerAction.kt
jeremymailen: You know I have a vague memory that they were stateful and if you didn't re-resolve them in the action you'd run into problems. Superstitiously I might recommend keeping the behavior. If you want to change it, make sure to run the PR version on a real project with a number of files and source sets.

jeremymailen:
I'm back from break and have some time to help get this one through. Thanks again!

src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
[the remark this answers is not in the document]
lwasyl: I did that, but seems like either some artifacts are still not available or the CI is flaky — 16/21 checks passed, and quick look at the logs shows stuff I'm not sure is related to this PR 🤔

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/format/FormatWorkerAction.kt
lwasyl: Hah, based on the documentation it looks like you're right in that the rules might be stateful, but at the same time `RuleSetProviderV2` solves statefulness issues (provided it's implemented as intended) and it's explicitly said that the engine won't reuse rules:

```kotlin
/**
 * The [Rule] contains the life cycle hooks which are needed for the KtLint rule engine.
 *
 * Implementation **doesn't** have to be thread-safe or stateless (provided that [RuleSetProviderV2] creates a new
 * instance of the [Rule] when [RuleSetProviderV2.getRuleProviders] has implemented its [RuleProvider] in such a way
 * that each call to [RuleProvider.createNewRuleInstance] indeed creates a new instance. The KtLint engine never
 * re-uses a [Rule] instance once is has been used for traversal of the AST of a file.
 */
public open class Rule(
```

It might be a good case to solve when adding tests involving custom rulesets though. I'll also check the current implementation on my $workProject 👍

src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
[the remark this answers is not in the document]
mateuszkwiecinski: Some of the checks failed due to code style offences:
image

the rest are caused by plugin's Gradle 8.0 incompatibility, which I'll try to address in: https://github.com/jeremymailen/kotlinter-gradle/pull/306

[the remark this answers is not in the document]
jeremymailen: #306 is merged now, so maybe try rebasing?

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/format/FormatWorkerAction.kt
jeremymailen: Yeah, they may well have fixed this issue in later ktlint versions -- so long as it tests out ok, I'm good with a change here 👍

src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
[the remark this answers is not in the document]
mateuszkwiecinski: The tests were still failing, so I reproed and reported here: https://github.com/pinterest/ktlint/issues/1768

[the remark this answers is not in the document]
lwasyl: Okay so, I think I've fixed the extension and I kept old behavior (using `EditorConfigOverride`, but after discussion in https://github.com/pinterest/ktlint/issues/1768 I'm somewhat convinced that Kotlinter should _not_ offer API to enable/disable rules, experimental or otherwise. And if it's really desireable, it shouldn't use `EditorConfigOverride` for that but rather `EditorConfigDefaults`.

Long explanation is here: https://github.com/paul-dingemans/ktlint/commit/cfecedcac8865d5a2c57466572ed387255e861a6

My take is that Ktlint respects hierarchical `.editorconfig`: you can override rules for each directory, while Kotlinter offers just a single global setting. The only reasonable behavior for Kotlinter is to provide its settings as a default rather than override, otherwise it's disallowing per-directory settings. But there are arguments to be made for removing Kotlinter API for configuring Ktlint completely:
 - users should have `.editorconfig` file in their project root directory even if they don't have any settings there, because without `root = true` Ktlint will search for `.editorconfig` outside of the project root directory (last I heard)
 - once there exists an  `.editorconfig` file, it's reasonable to have it as a central source of configuration. Having global defaults/overrides in Kotlinter seems like a small benefit. Perhaps the API was added when it was more difficult (or impossible) to use `.editorconfig`, but clearly Ktlint considers it a sole source of configuration now and I think it's best not to fight it

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/lint/LintWorkerAction.kt
```diff
@@ -31,28 +30,30 @@ abstract class LintWorkerAction : WorkAction<LintWorkerParameters> {
     private val ktLintParams: KtLintParams = parameters.ktLintParams.get()
 
     override fun execute() {
-        resetEditorconfigCacheIfNeeded(
-            changedEditorconfigFiles = parameters.changedEditorconfigFiles,
+        val ktLintEngine = KtLintRuleEngine(
+            ruleProviders = resolveRuleProviders(defaultRuleSetProviders, ktLintParams.experimentalRules),
```
mateuszkwiecinski, line 34: ❓ 
Since experimental rules can be now also enabled via the editorconfig, shouldn't we always append them here? 
```suggestion
            ruleProviders = resolveRuleProviders(defaultRuleSetProviders, true),
```
I imagine it would be surprising to the users if they enabled experimental rules via `.editorconfig`, as stated in ktlint docs, and the rules wouldn't be used out-of-the-box.

src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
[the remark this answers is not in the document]
mateuszkwiecinski: I fully agree 👍  I'd strongly suggest moving away from `EditorConfigOverride` in favour of `EditorConfigDefaults`, as [Paul suggested ](https://github.com/pinterest/ktlint/issues/1768#issuecomment-1383196514)- it should be less confusing for the user.

Actually, I'd even consider taking a step further, and deprecate both `disabled_rules` and `experimentalRules` within the plugin, since it probably should be preferred to pass them via `.editorconfig` properties (wouldn't it?). My main motivation here is to get rid of some of the complexity we're adding, just to preserve backwards compatibility. Moreover, AFAICS configuring rules via `.editorconfig` requires that same amount of effort as via Plugin's extension, so the extension doesn't provide that much of a value anymore.
This might require a separate discussion though 👀

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/lint/LintWorkerAction.kt
lwasyl: So just to be clear, you're suggesting to change the plugin behavior for this version? I aimed at keeping compatibility and not introducing any new changes, but passing `true` here (in fact I can just remove this param whatsoever) _will_ change the behavior 🤔 I thought it was possible to enable experimental rules already before, no?

I'd wait for @jeremymailen input here, I agree with you that we should load all rules, but at the same time I feel like doing that without moving to `EditorConfigDefaults` or deprecating Kotlinter rules configuration is a half-solution and might end up being more confusing. As a user, I'd prefer to have a single release which fully revamps the way I configure the rules

mateuszkwiecinski: >So just to be clear, you're suggesting to change the plugin behavior for this version? 

Not really 🤔 I mean, the plugin is just a wrapper around ktlint. It avoids overriding default ktlint behavior. So if ktlint now allows enabling experimental rules via `editorconfig`, I don't see a reason why we would want to prevent that.

> I thought it was possible to enable experimental rules already before, no?

As far as I understand, before 0.48.0, the only way to enable experimental rules was to pass additional RuleSet providers, which had to be done by the plugin. Now, as a ktlint user, by adding the `ktlint_experimental=true`, I'd expect the experimental rules to be enabled.
Screenshot from their [documentation](https://pinterest.github.io/ktlint/faq/#why-is-editorconfig-property-disabled_rules-deprecated-and-how-do-i-resolve-this)
image


> but passing true here will change the behavior

Which behavior will change specifically? 🤔 Starting from 0.48.0 ktlint won't run experimental rules unles they are explicitly enabled, right?

lwasyl: Ah I thought it was possible to enable experimental rules via `.editorconfig` before as well. So yes, pulling all rules should not be harmful, but it will still be confusing if someone enables experimental rules in Kotlinter but wants to disable them only for some directory. We'd support _some_ Ktlint configuration but not fully, it _can_ get confusing

lwasyl: Done, all rules will now be passed to Ktlint

src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
```diff
@@ -1,31 +1,75 @@
...
+    if (ktLintParams.experimentalRules) {
+        listOf(
+            EditorConfigProperty(
+                type = PropertyType("ktlint_experimental", "Experimental rules", IDENTITY_VALUE_PARSER),
+                defaultValue = "enabled",
+            ) to "enabled",
+        )
+    } else {
+        emptyList()
+    }
```
mateuszkwiecinski, line 49: 🔧 One last thing:
Since the goal of this PR is to support ktlint >=0.48, and they introduced a feature where you can hierarchically configure experimental rules via `.editorconfig` integration, I would strongly suggest to not block this behavior. With current state, if user enables experimental rules via plugin extension they won't be able to disable them via `.editorconfig` (i.e. for a specific directory). Having the "global" setting used as the default sounds much more reasonable to me. 

Keeping scope of this PR small - I'd suggest keeping the `editorConfigOverride` for disabled rules (to preserve backwards compatibility), but start using `editorConfigDefaults` for experimental rules (since that's a new thing we're adding right now).

mateuszkwiecinski: ❔ 
And if the decision is to give plugin's extension priority, shouldn't we then _force_ the disabled state as well? To serve a _symmetrical_ behavior
```suggestion
private fun getPropertiesForExperimentalRules(ktLintParams: KtLintParams) =
	listOf(
	    EditorConfigProperty(
	        type = PropertyType("ktlint_experimental", "Experimental rules", IDENTITY_VALUE_PARSER),
	        defaultValue = "enabled",
	    ) to if(ktLintParams.experimentalRules) "enabled" to "disabled",
	)
```

mateuszkwiecinski (commented):
One last comment, the rest LGTM 🚀

mateuszkwiecinski (approved):

src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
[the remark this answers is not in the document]
jeremymailen: Yes, I agree. We should stop configuring rules altogether in our gradle config block. The config block remains but just for kotlinter specific gradle configurations such as ignore failures and reporters. A few consequences:

- Our next kotlinter release will be 4.x compatibility breaking
- We can include the other compatibility breaking changes in this release #279 and #294 
- Sounds like we need not wait for ktlint 0.48.2 since we don't need that editorconfig fix anymore?

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/lint/LintWorkerAction.kt
jeremymailen: Yeah, as mentioned above, I support removing kotlinter's configuration of rules and delegating this entirely to editorconfig and ktlint's use of it. I couple assumptions I'm making
- This harmonizes with people adding customer rule set to the build? I think so, just mentioning it
- We properly set the gradle targets as outdated when various layers of editorconfig change -- I think we got this change last time around with the enhanced ktlint APIs if I remember right.

src/main/kotlin/org/jmailen/gradle/kotlinter/support/EditorConfigUtils.kt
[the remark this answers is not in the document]
lwasyl: > Sounds like we need not wait for ktlint 0.48.2 since we don't need that editorconfig fix anymore?

Looks like it to me 🙂 Even with current implementation we don't need 0.48.2 I think, only 0.48.0 was breaking Kotlinter

@jeremymailen Please let me know how you'd like to proceed with this PR. Personally I'd prefer to merge it as is, as it's been open for a while now. Removing/deprecating rules configuration will probably involve writing some docs so I'd prefer to leave that to you (or at least a separate review)

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/lint/LintWorkerAction.kt
lwasyl: > This harmonizes with people adding customer rule set to the build? I think so, just mentioning it

Afaiu yes, the rules will be _added_ via Kotlinter mechanism and configured in `.editorconfig`. Though now that I think of it, I guess Ktlint might even want to load rules on the classpath itself

> We properly set the gradle targets as outdated when various layers of editorconfig change

That I don't fully understand so can't answer

jeremymailen (approved):
Yes, please proceed. We'll remove the rule configuration in a separate change.

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/format/FormatWorkerAction.kt
jeremymailen: Actually, discovered you do in fact need to resolve rule providers for each file. If you don't, you get an error when KtLint tries to parse the .editorconfig on rule like
```
ktlint_standard_no-wildcard-imports = disabled
```

It will complain that ruleSetExecution is null in the VisitorProvider.

I'm putting together a 3.14.0 release and I'll fix there.

Issue #320, opened 2023-03-20
Compatibility matrix is out of date


rsmith20:
The compatibility matrix specifies 3.11+ as compatible with min gradle version 7.0, but 3.14 is compiled using Kotlin 1.8 and gradle 7.5 expects Kotlin 1.6.

jeremymailen:
So you are saying the new row needs to be 3.14+ compatible with gradle 7.6+?

mateuszkwiecinski:
I'd suggest not updating the compatibility in readme, but rather revert [these two](https://github.com/jeremymailen/kotlinter-gradle/blob/5c890e84bcc1b19bbaf92f5b50a9505c4b3e3807/build.gradle.kts#L101-L102) properties back to 1.4. 
```
kotlinOptions {
    apiVersion = "1.4"
    languageVersion = "1.4"
}
```
This should restore plugin compatibility with lower versions. Gradle enforces Kotlin version in runtime, so these settings effectively are not respected and only cause incompatibilities 🤷 

The runtime with Kotlin 1.8 is available only on Gradle 8: https://docs.gradle.org/current/userguide/compatibility.html#kotlin

jeremymailen:
I bumped it because IntelliJ indicated that 1.4 was deprecated.
Is there significant user base still using a version of gradle that old?

jeremymailen:
I think the compatibility matrix is up-to-date, but please reopen if there are still issue.

PR #330, opened 2023-05-25
Upgrade to ktlint-0.49.1

jeremymailen:
A few twists and turns in the migration, but functionality unaffected for the most part.

Also
- Upgrade Kotlin and Gradle
- Fix docs issue #320

### DIFF
--- a/src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/lint/LintWorkerAction.kt
+++ b/src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/lint/LintWorkerAction.kt
@@ -1,14 +1,16 @@
 package org.jmailen.gradle.kotlinter.tasks.lint
 
-import com.pinterest.ktlint.core.Code
-import com.pinterest.ktlint.core.Reporter
+import com.pinterest.ktlint.cli.reporter.core.api.KtlintCliError
+import com.pinterest.ktlint.cli.reporter.core.api.ReporterV2
+import com.pinterest.ktlint.rule.engine.api.Code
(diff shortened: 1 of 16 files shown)


src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/lint/LintWorkerAction.kt
```diff
@@ -45,14 +46,14 @@ abstract class LintWorkerAction : WorkAction<LintWorkerParameters> {
                     return@forEach
                 }
 
-                ktLintEngine.lint(Code.CodeFile(file)) { error ->
+                ktlintEngine.lint(Code.fromFile(file)) { error: LintError ->
```
jeremymailen, line 49: To me this is a bit baffling. `lint` api calls return `LintError` which seems legacy, but we have to feed a `KtLintCliError`, which is slightly richer, into Reporters. The name almost sounds like it's being moved out of API consumer land?
In any case, made it work.

src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/lint/LintWorkerAction.kt
```diff
@@ -68,3 +69,15 @@ abstract class LintWorkerAction : WorkAction<LintWorkerParameters> {
 }
 
 private val supportedExtensions = setOf("kt", "kts")
+
+internal fun LintError.toCliError() = KtlintCliError(
+    line,
+    col,
+    ruleId.value,
+    detail,
+    if (canBeAutoCorrected) {
```
jeremymailen, line 78: And this seems like a lossy conversion

PR #357, opened 2023-12-12
Fix lint warning

marukami:
Replace buildDir with layout.buildDirectory.


jeremymailen (approved):
Thank you for the fix.

PR #364, opened 2024-03-15
Update README.md

realdadfish:
Easier ktlint version override utilizing the provided bom.
Also bump the KtLint version to 1.2.1, as 4.2.0 already utilizes KtLint 1.1.1 and the example was misleading.


jeremymailen (approved):
Thank you!
Looks good. Somehow I didn't know they had a bom.

PR #420, opened 2024-12-08
Run Kotlinter in process isolation

jeremymailen:
First in a series of changes leading to a 5.0.0 release. Starting with Kotlin 2.1.0 it's no longer acceptable to run ktlint and the embedded Kotlin compiler in the main Gradle classloader. The solution is to use Gradle worker process isolation.

Positive consequences:
- Users can configure the version of ktlint
- Kotlinter is compatible with any version of Kotlin with matching language specification

Side effects to watch:
- Performance may be slower

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -20,7 +20,7 @@ val githubUrl = "https://github.com/jeremymailen/kotlinter-gradle"
 val webUrl = "https://github.com/jeremymailen/kotlinter-gradle"
 val projectDescription = "Lint and formatting for Kotlin using ktlint with configuration-free setup on JVM and Android projects"
 
-version = "4.5.0"
+version = "5.0.0-M1"
 group = "org.jmailen.gradle"
 description = projectDescription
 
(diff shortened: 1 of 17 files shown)


jeremymailen:
Tested it on a substantial multi-module project with command `gradle clean lintKotlin`:
- Before: 13s
- After: 25s

So it will be slower, but this is the cost of isolation. Incremental of course is still fast and under 1s because generally not many files change between runs.

jeremymailen:
We do experience the [classloader leak bug](https://github.com/gradle/gradle/issues/18313) however.
If I configure gradle.properties:
```
org.gradle.jvmargs=-Xmx1024m "-XX:MaxMetaspaceSize=1024m"
```
This helps quite a bit, but even then if you repeat `clean lintKotlin` several times it will run the gradle daemon out of metaspace. It may reclaim memory, but not aggressively enough.

@osipxd you seem to have some insight here. Is the workaround to use process isolation instead of classloader? It seems like we could run into different problems there.

osipxd:
I haven't investigated possible workarounds deeply myself. But I've seen another possible workaround by @martinbonnin:
- https://github.com/apollographql/apollo-kotlin/pull/6305

jeremymailen:
> Tested it on a substantial multi-module project with command `gradle clean lintKotlin`:
> 
> * Before: 13s
> * After: 25s
> 
> So it will be slower, but this is the cost of isolation. Incremental of course is still fast and under 1s because generally not many files change between runs.

Switching to process isolation yields:
- After: 19s

So it's actually better than classloader isolation, but still a small performance hit.

martinbonnin:
FWIW, [we went the process isolation route](https://github.com/apollographql/apollo-kotlin/pull/6267) before but that created other questions like how many processes can you run in parallel? What heap size to allocate in each process, etc... It can work (and from the issues linked in https://github.com/gradle/gradle/issues/18313, looks like this is definitely a solution) but it's no silver bullet either.

martinbonnin:
FYI, Gradle has a fix 🎉 : https://github.com/gradle/gradle/pull/31654

OliverO2:
FYI: Fix seems to appear in Gradle 9.8 via https://github.com/gradle/gradle/pull/38527.

jeremymailen:
> FYI: Fix seems to appear in Gradle 9.8 via [gradle/gradle#38527](https://github.com/gradle/gradle/pull/38527).

This is good news. I suppose for backwards compatibility with other versions of gradle we might need to sense the gradle version or provide a configurable mode to allow people to select.

PR #553, opened 2026-07-31
Suppress JDK 24+ sun.misc.Unsafe warning from the ktlint worker

### LINKED ISSUE #552: Lint/Format worker prints JDK 24+ sun.misc.Unsafe warning with no way to suppress it
## Summary

On JDK 24+, every build that runs `lintKotlin`/`formatKotlin` prints:

```
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::objectFieldOffset has been called by org.jetbrains.kotlin.com.intellij.util.containers.Unsafe (file:.../kotlin-compiler-embeddable-2.2.21.jar)
WARNING: Please consider reporting this to the maintainers of class org.jetbrains.kotlin.com.intellij.util.containers.Unsafe
WARNING: sun.misc.Unsafe::objectFieldOffset will be removed in a future release
```

Per [JEP 498](https://openjdk.org/jeps/498) this is suppressed with `--sun-misc-unsafe-memory-access=allow`.
There is currently **no way for a build author to get that flag to the ktlint worker JVM.**

## Environment

- kotlinter 5.5.0 (bundles ktlint 1.8.0)
- Gradle 9.6.1, Kotlin plugin 2.4.0
- JDK 25.0.4

## Why the bundled ktlint version doesn't help

ktlint already handled this in [ktlint#2973](https://github.com/ktlint/ktlint/issues/2973) via
[PR#3040](https://github.com/ktlint/ktlint/pull/3040) - but only by adding the flag to the **CLI
launcher scripts**. kotlinter 5.5.0 bundles ktlint 1.8.0, so the fix is present yet unreachable,
because the plugin invokes the engine in a worker rather than through the CLI launcher.

ktlint's maintainer noted upstream that ktlint itself can't fix the root cause, since it needs the
embedded Kotlin compiler. ktlint's own
[integrations docs](https://github.com/ktlint/ktlint/blob/master/documentation/release-latest/docs/install/integrations.md)
recommend `jvmArgs("--sun-misc-unsafe-memory-access=allow")` for `JavaExec`-based integrations.

## Suggested fix

Set the flag on the worker's fork options when running on JDK 24+:

```kotlin
val workQueue = workerExecutor.processIsolation { config ->
    config.classpath.setFrom(ktlintClasspath)
    if (JavaVersion.current() >= JavaVersion.VERSION_24) {
        // Suppress "sun.misc.Unsafe::objectFieldOffset" warning from the embedded Kotlin compiler
        config.forkOptions { it.jvmArgs("--sun-misc-unsafe-memory-access=allow") }
    }
}
```

`ProcessWorkerSpec` extends `ForkingWorkerSpec`, which provides
`forkOptions(Action<? super JavaForkOptions>)`, so this needs no new Gradle API.

Alternatively (or additionally), exposing the worker's fork options / a `jvmArgs` property on
`ConfigurableKtLintTask` would let build authors handle this and similar cases themselves.

szymonsasin-nordic:
Fixes #552.

The Kotlin compiler embedded in ktlint calls `sun.misc.Unsafe::objectFieldOffset`, so on JDK 24+ every `lintKotlin`/`formatKotlin` run prints:

```
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::objectFieldOffset has been called by org.jetbrains.kotlin.com.intellij.util.containers.Unsafe (...)
```

ktlint addressed this in [ktlint#2973](https://github.com/ktlint/ktlint/issues/2973) by adding `--sun-misc-unsafe-memory-access=allow` to its CLI launcher scripts, which the worker never goes through, so the bundled fix is unreachable here.

## Changes

- `--sun-misc-unsafe-memory-access=allow` is passed to the ktlint worker JVM on JDK 24+ ([JEP 498](https://openjdk.org/jeps/498)).
- New `workerJvmArgs: ListProperty<String>` on `ConfigurableKtLintTask`, holding that default. `add` keeps it, `set` replaces it — so build authors can pass their own arguments, and can drop the default themselves once a future JDK stops accepting the option, without waiting for a release.
- `LintTask` and `FormatTask` now share one `ktlintWorkQueue()` helper instead of each configuring process isolation.

Two notes on the implementation:

- The version check uses `Runtime.version().feature()` rather than `JavaVersion.VERSION_24`. That enum constant only exists in Gradle versions that already knew about JDK 24, and the README documents support back to Gradle 8.4.
- `workerJvmArgs` is `@Internal`. JVM arguments don't affect lint results, and as an `@Input` they would make the cache key of the `@CacheableTask` `LintTask` depend on the local JDK version, so machines on JDK 21 and JDK 25 would stop sharing build cache entries.

## Testing

- Unit test for the version gate (24/25 → flag, 11/21 → no arguments).
- Functional tests asserting `lintKotlin` output contains no `sun.misc.Unsafe` warning, and that a custom `workerJvmArgs` value reaches the worker JVM.
- Verified by hand in `test-project` on JDK 25: the warning appears before the change and is gone after, for both `lintKotlin` and `formatKotlin`. `./gradlew check` passes, and `test-project` also runs clean with the daemon on JDK 21 (no-flag path).


jeremymailen:
Thank you! Will review this weekend.

jeremymailen (approved):
Thank you for the excellent and thorough contribution! 🙌

